PR #23, opened 2019-08-02
Add findstructralnz(A,col_index) for CSC and BBB matrix

huanglangwen:
Indexing over BandedBlockBandedMatrix in colored jacobian (https://github.com/JuliaDiffEq/DiffEqDiffTools.jl/issues/67) is much slower than CSC matrix (2.5ms vs 120ms). I changed it into a column based iteration according to https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/483 . The evaluation time of colored jacobian for BBB matrix is reduced to 87ms.

### DIFF
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -96,6 +96,18 @@ has_sparsestruct(x::Type{<:Tridiagonal}) = true
 has_sparsestruct(x::Type{<:SymTridiagonal}) = true
 
 """
+    fast_column_indexing(x::AbstractArray)
+
+determine whether indexing along columns is fast, if it is `true`
+`findstructralnz(x,col_index)` is used, otherwise `findstructralnz(x)`
+is used.
+"""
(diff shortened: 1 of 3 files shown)


src/ArrayInterface.jl
```diff
@@ -95,6 +95,18 @@ has_sparsestruct(x::Type{<:Bidiagonal}) = true
...
+"""
+    fast_column_indexing(x::AbstractArray)
+
+determine whether indexing along columns is fast, if it is `true`
+`findstructralnz(x,col_index)` is used, otherwise `findstructralnz(x)`
+is used.
+"""
+fast_column_indexing(x) = false
+fast_column_indexing(x::AbstractArray) = fast_column_indexing(typeof(x))
+fast_column_indexing(x::Type{<:AbstractArray}) = false
```
ChrisRackauckas: it should definitely be true for Array

PR #26, opened 2019-11-13
Mark AbstractQ and LQPackedQ as non-fast_scalar_indexing

tkf:
Their `getindex` are defined in terms of multiplications.  So I guess it makes sense to mark them as non-`fast_scalar_indexing`?

```julia
function getindex(Q::AbstractQ, i::Integer, j::Integer)
    x = zeros(eltype(Q), size(Q, 1))
    x[i] = 1
    y = zeros(eltype(Q), size(Q, 2))
    y[j] = 1
    return dot(x, lmul!(Q, y))
end
```

--- https://github.com/JuliaLang/julia/blob/6eebbbe2d205f8116330a77ca5e15f4a356232db/stdlib/LinearAlgebra/src/qr.jl#L526-L532

```julia
getindex(A::LQPackedQ, i::Integer, j::Integer) =
    lmul!(A, setindex!(zeros(eltype(A), size(A, 2)), 1, j))[i]
```

--- https://github.com/JuliaLang/julia/blob/6eebbbe2d205f8116330a77ca5e15f4a356232db/stdlib/LinearAlgebra/src/lq.jl#L139-L140

### DIFF
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -50,6 +50,8 @@ Query whether an array type has fast scalar indexing
 """
 fast_scalar_indexing(x) = true
 fast_scalar_indexing(x::AbstractArray) = fast_scalar_indexing(typeof(x))
+fast_scalar_indexing(::Type{<:LinearAlgebra.AbstractQ}) = false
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Yes, good point!

PR #28, opened 2019-11-15
Default to ismutable(array) = true

tkf:
I propose to get rid of the default `ismutable(x::DataType) = x.mutable`.  An array type implemented with immutable struct does not mean the array is immutable (e.g., sparse arrays #27).  Like wise, it is imaginable to implement semantically immutable array using mutable struct.

So, I think it would be nice to have a sane and predictable default.  Here, I suggest to treat arrays as mutable by default.  This way, missing definition of `ismutable` is an error, rather than a slow or incorrect computation.


tkf:
Thanks, good to know that you are OK with the change.  After this is merged, can you release the package soon?  I'd like to use it from https://github.com/tkf/BangBang.jl/pull/48

PR #32, opened 2019-12-23
Add appropriate interface for ranges

Tokazama:
Only changes are to `ismutable` and `can_setindex` for `AbstractRange` types.


ChrisRackauckas:
Thanks!

PR #33, opened 2019-12-26
Update for new BlockArrays

dlfivefifty:


dlfivefifty:
Ah, I've dropped support for Julia 1.0. It might be better to just cap the BlockBandedMatrices.jl version to 0.6. (I'm not sure how you do this with extras)

ChrisRackauckas:
Might as well increase the lower bounds here. Thanks for the update!

PR #36, opened 2020-02-02
general function for AoS -> SoA

ChrisRackauckas:
Since this will likely be needed in a lot more places.

### DIFF
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -45,6 +45,13 @@ can_setindex(x::AbstractArray) = can_setindex(typeof(x))
 can_setindex(::Type{<:AbstractRange}) = false
 
 """
+    aos_to_soa(x)
+
+Converts an array of structs formulation to a struct of array
+"""
+aos_to_soa(x) = x
+
+"""
     fast_scalar_indexing(x)
 
 Query whether an array type has fast scalar indexing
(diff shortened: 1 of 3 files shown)


src/ArrayInterface.jl
```diff
@@ -44,6 +44,13 @@ can_setindex(x) = true
 can_setindex(x::AbstractArray) = can_setindex(typeof(x))
 can_setindex(::Type{<:AbstractRange}) = false
 
+"""
+    aos_to_soa(x)
+
+Converts an array of structs formulation to a struct of array
+"""
+aos_to_soa(x) = x
```
matbesancon: I'm a bit late on this, but `aos_to_soa` does not read that well, two not-that-known abbreviations in an exported function

ChrisRackauckas: Yeah, but anything I could think of was excessively long, so 🤷‍♂

matbesancon: `structarray_to_arraystruct`? `rowarray_to_colarray`? There should be a similar name in the data ecosystems

ChrisRackauckas: `rowarray_to_colarray` doesn't work since it really is more general structs. `structarray_to_arraystruct` would be possible.

matbesancon: alternatively, writing one of them in full letter and the second in abbreviation also works, since they are redundant, `structarray_to_aos`?

ChrisRackauckas: ehh I like the symmetry

PR #39, opened 2020-02-28
Add `lu_instance`

YingboMa:
I am not sure if `lu_object` is a good name.

Ref: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/1049


ChrisRackauckas:
`lu_instance`?

ChrisRackauckas:
Add a StaticArrays one too?

PR #41, opened 2020-03-10
Add ismutable(tuple) = false

ChrisRackauckas:


src/ArrayInterface.jl
```diff
@@ -12,11 +12,12 @@ function ismutable end
 Query whether a type is mutable or not, see
 https://github.com/JuliaDiffEq/RecursiveArrayTools.jl/issues/19.
 """
-ismutable(x) = ismutable(typeof(x))
+ismutable(x) = (@show x; ismutable(typeof(x)))
```
mauro3: reomote `@show`

ChrisRackauckas: thanks!

ChrisRackauckas:
```suggestion
ismutable(x) = ismutable(typeof(x))
```

PR #42, opened 2020-03-18
add issingular

ChrisRackauckas:
@yingboma do you have ideas for other dispatches that would be good? Should there be a fallback?

### DIFF
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -114,6 +114,10 @@ Determine whether a given abstract matrix is singular.
 issingular(A::Matrix) = !issuccess(lu(A, check=false))
 issingular(A::UniformScaling) = A.λ == 0
 issingular(A::Diagonal) = any(iszero,A.diag)
+issingular(A::Union{Hermitian,Symmetric}) = diaganyzero(bunchkaufman(A, check=false).LD)
+issingular(A::Union{LowerTriangular,UpperTriangular}) = diaganyzero(A.data)
+issingular(A::Union{UnitLowerTriangular,UnitUpperTriangular}) = false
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Can you take it from here? It seems you have a bunch of ideas so that might be quicker.

YingboMa:
Done. @ChrisRackauckas could you take a look?

src/ArrayInterface.jl
```diff
@@ -106,6 +106,19 @@ has_sparsestruct(x::Type{<:Bidiagonal}) = true
...
 has_sparsestruct(x::Type{<:SymTridiagonal}) = true
 
+"""
+    issingular(A::AbstractMatrix)
+
+Determine whether a given abstract matrix is singular.
+"""
+issingular(A::Matrix) = !issuccess(lu(A, check=false))
+issingular(A::UniformScaling) = A.λ == 0
+issingular(A::Diagonal) = any(iszero,A.diag)
```
dkarrasch: One could add here
```julia
issingular(B::Bidiagonal) = any(iszero, A.dv)
issingular(S::SymTridiagonal) = diaganyzero(iszero, ldlt(S).data)
issingular(T::Tridiagonal) = !issuccess(lu(A, check=false))
```
And perhaps as a fallback (which fulfills the promise of the docstring)
```julia
issingular(A::AbstractMatrix) = issingular(Matrix(A))
issingular(A::AbstractSparseMatrix) = !issuccess(lu(A, check=false))
```

ChrisRackauckas:
Note I reverted the README change @YingboMa but it's expected that `issingular` is the best way to check for it: if it's ever faster to use `rank`, then the dispatch should be to use `rank`.

PR #52, opened 2020-05-26
Add `parent_type` method

Tokazama:
This adds a method that returns the type of the parent array. The default method is to just return the type of the array itself, which matches `parent`'s behavior.

If this sounds good I'll move forward with writing tests.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -164,7 +164,7 @@ using ArrayInterface: zeromatrix
 @test zeromatrix(rand(4,4,4)) == zeros(4*4*4,4*4*4)
 
 using ArrayInterface: parent_type
-@testset "" begin
+@testset "Parent Type" begin
     x = ones(4, 4)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
This looks great. Would be good to have tests. I find myself wanting this often.

test/runtests.jl
```diff
@@ -162,3 +162,14 @@ end
 
 using ArrayInterface: zeromatrix
 @test zeromatrix(rand(4,4,4)) == zeros(4*4*4,4*4*4)
+
+using ArrayInterface: parent_type
+@testset "" begin
```
ChrisRackauckas, line 167:
```suggestion
@testset "Parent Type" begin
```

ChrisRackauckas:
Oh, forgot to add it to the README. Follow up?

ChrisRackauckas:
Actually, isn't an easier implementation just `typeof(parent(A))`?

PR #55, opened 2020-06-25
ComponentArray compatibility

YingboMa:


src/ArrayInterface.jl
```diff
@@ -501,6 +501,10 @@ function __init__()
                                                                                       0)
   end
 
+  @require ComponentArrays="b0b7db55-cfe3-40fc-9ded-d10e2dbeff66" begin
+    lu_instance(jac_prototype::ComponentArrays.ComponentArray) = lu_instance(ComponentArrays.getdata(jac_prototype))
```
ChrisRackauckas, line 505: we should really be having people take this as a dependency, but it's fine for now.

Issue #50, opened 2020-04-17
AbstractRange traits


Tokazama:
Would it be within the scope of this package to address range traits? I have several ideas that I figured would be more useful here than in a separate package.

ChrisRackauckas:
Yes, that could be useful here as well.

Tokazama:
The only trait in base we have is `RangeStepStyle` which isn't extensively used and has variable interpretations (see https://github.com/JuliaLang/julia/pull/35279).

It may also be useful to have traits that are based on how a range is constructed. For example, there's variability in how the start of a range is determined (e.g., `Base.Oneto` always begins at one, `UnitRange` and `StepRange` always begin at `range.start`, `StepRangeLen` is derived from a reference value and an offset). There is similar variability in how the last element of a range is determined. These would be helpful for creating fast look up/indexing methods without worrying about the type hierarchy of of `AbstractRange` as much.

chriselrod:
What about something for extracting static/known size information?
Something like:
```julia
known_start(::Any) = nothing
known_stop(::Any) = nothing
known_start(::Type{<:Base.OneTo}) = 1

# As an example, StaticArrays could add support via:
using StaticArrays
known_start(::Type{<:StaticArrays.SOneTo}) = 1
known_stop(::Type{<:StaticArrays.SOneTo{N}}) where {N} = N
```

Tokazama:
@chriselrod I like this approach a lot. It would also be worth having

```julia
known_step(::Any) = nothing
known_step(::Type{<:AbstractUnitRange{T}}) where {T} = one(T)
```

Just these three simple traits would make it a lot easier to write generic methods for ranges.

ChrisRackauckas:
I agree.

PR #56, opened 2020-07-18
Added sized range interface.

chriselrod:
As brought up in https://github.com/SciML/ArrayInterface.jl/issues/50

Anything else we should add while at it?

Thoughts on the name and default return (`nothing`)?
I figured `known_` is clear as a prefix for queering values known at compile time.


ChrisRackauckas:
>Anything else we should add while at it?

Add the definition to the README

Issue #22, opened 2019-08-01
Add an `is_resizable` trait?


rofinn:
It'd be nice if there was an abstract array specifying whether an array is resizable in-place.

ChrisRackauckas:
Agreed.

oxinabox:
Is it worth distighishing this into:
`isgrowable` for things that can increase size (like `Vector`)
and
`isresizable` that can just shuffle the size around into different dimensions?

Or maybe it is a trait with multiple values:
 - `Growable()` > `Resizable` > `FixedSize` ?

oxinabox:
Really i feel like we want to answer questions about:
 
 - `hasmethod(resize!, ...)`
 - `hasmethod(push!, ...)`  (and maybe its flavours `pushfirst!`)
 - `hasmethod(deleteat, ...)` (and maybe its flavors `pop!`, `popfirst`, `filter!`)
 - maybe `hasmethod(reshape,...)` ?

ChrisRackauckas:
If `hasmethod` is a compile-time "trait", then do we still need a trait for this?

oxinabox:
Yes, to group them. And to name them. But anyway Jameson says my plan is impossible.

Tokazama:
I created something similar for this here https://github.com/Tokazama/StaticRanges.jl/blob/f8dac65496c10ea0ec0fbf748006af105f77f62e/src/staticness.jl#L1.
It separates things into dynamic, static, fixed. Where a `Vector` is dynamic, a typical range is fixed, and `SVector` is static. Would this be a good solution to this?

ChrisRackauckas:
an `isresizable` trait function is all that's needed.

tkf:
FYI, BangBang internally now has [`implements(f!, x)`](https://tkf.github.io/BangBang.jl/dev/internals/#BangBang.implements-Tuple{Any,Any}) and [`possible(f!, args...)`](https://tkf.github.io/BangBang.jl/dev/internals/#BangBang.possible) to query the capability of containers/objects.  ATM only `implements(setindex!, x)`, `implements(push!, x)`,  and `implements(setproperty!, x)` are defined.  They are like `hasmethod` but manually implemented (like @oxinabox mentioned https://github.com/JuliaDiffEq/ArrayInterface.jl/issues/22#issuecomment-518005555).  I also need to consider "element" types so that's why there is `possible` function that can do:

```julia
julia> possible(push!, Int[], 1)
true

[...]

PR #60, opened 2020-08-11
New trait for collections that change sizes (issue #22)

Tokazama:
Even if you have a mutable structure it might not be able to change sizes. With this PR if `is_dynamic` returns `true` then the following operations should work for a given collection:

* `push!`
* `pushfirst!`
* `pop!`
* `popfirst!`
* `insert!`
* `deleteat!`

I also updated the README with these methods and can bump the version number if this all sounds good.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,9 +15,9 @@ and exported in a future Base Julia there will be no issues with the upgrade.
 
 Returns the parent array that `x` wraps.
 
-## is_dynamic(x)
+## can_change_size(x)
 
-Returns `true` if the size of `T` is dynamic. If `T` is dynamic then operations
+Returns `true` if the size of `T` can change, in which case operations
 such as `pop!` and `popfirst!` are available for collections of type `T`.
 
 ## ismutable(x)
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
`is_dynamic_sized`?

chriselrod:
I'm a little concerned about ambiguity of dynamic referring to changing size vs size-unknown-at-compile time.

Tokazama:
> I'm a little concerned about ambiguity of dynamic referring to changing size vs size-unknown-at-compile time.

I did bring up having a distinction between static, fixed, and dynamic sizes in the referenced issue but I didn't include these because something like `UnitRange` couldn't have a dynamic size even if it's not known at compile time. I'd be happy to include them if it's seen as a necessary inclusion.

> `is_dynamic_sized` ?

Although that's more specific I'm not sure what else would be "dynamic", so I prefer the `is_dynamic` syntax. However, I'd be open to changing if people feel strongly.

ChrisRackauckas:
`can_change_size`

Tokazama:
`can_change_size` seems like it would be much more consistent with the other methods here. I'll change it.

PR #61, opened 2020-08-11
Added some methods to help support LoopVectorization.jl

chriselrod:
The methods are:
- `canavx(f)`, a function whitelist for supporting `@avx`.
- `is_cpu_column_major`. The name should be explanatory; figured it'd be best to go with something unambiguous.
- `stridelayout`. Provides a description of the memory layout.

Tagging @LaurentPlagne, we we discussed the intention of `stridelayout` earlier.
`stridelayout` is meant to decouple the axis that is good for SIMD (indicated by `contig`) from cache-friendliness (lower `rank`s). Our goal additionally is to decouple these from the array's interface, so that we can easily optimize the memory layouts of our arrays without having to touch any of the code.
This is facilitated by LoopVectorization, which will adjust the order of loops as necessary, making our code generic yet still optimal with respect to memory layout.

If you think of a `Vector{ComplexF64}` as a `2`x`N` `Matrix{Float64}`, then a `StructOfArray`s is essentially just a `transpose`, to allow you to interact with the array as before even though the memory layout is now essentially a `N`x`2` `Matrix{Float64}`, which should generally be much friendlier to SIMD.
However, the memory accesses for `re` and `im` numbers may now be far apart. While unlikely to be a problem in this example, this isn't necessarily the case for more general transformations of large arbitrary-dimensional arrays.
In this example, `contig,batch=2,4` yield a memory layout of `4`x`2`x`N/4`. That is, with the `AVX` instruction set, we could easily SIMD-load 4 contiguous elements of real and imaginary numbers at a time, yet these subsequent loads would still be contiguous and thus likely to trigger the DCU hardware prefetcher, even in more complicated examples.
As another example, for AVX-matrix multiplication `A * B` with contiguous-access packing, our `m_c x k_c` packed blocks of `A` would be reshaped into `8`x`k_c`x`m_c/8`. That is, it would correspond to `stridelayout(packed_A) == (1,8,(2,1))`: `m_c` is contiguous with batches of size `8`, but has a higher stride than `k_c`.

Being able to simply try out a bunch of different memory layouts without needing to change the code for correctness (thanks to API-memory layout decoupling) or performance (thanks to LoopVectorization's loop-reordering) should make this critical aspect of performance optimization much simpler, and even open the door to automation.


README.md
```diff
@@ -112,6 +112,21 @@ If `step` of instances of type `T` are known at compile time, return that step.
 Otherwise, returns `nothing`. For example, `known_step(UnitRange{Int})` returns
 `one(Int)`.
 
+## is_cpu_column_major(::Type{T})
```
ChrisRackauckas, line 115: why cpu?

README.md
```diff
@@ -112,6 +112,21 @@ If `step` of instances of type `T` are known at compile time, return that step.
...
+`pointer`. Returns `false` otherwise.
+
+## stridelayout(::Type{T})
+
+Returns a 3-tuple describing the strided layout of the memory of an instance of type `T` if it
+is known, and returning `nothing` otherwise.
+The elements of the tuple include
+ - `contig`: The axis with contiguous elements. `contig == -1` indicates no axis is contiguous. `striderank[contig]` does not necessarilly equal `1`.
+ - `batch`: indicates the number of contiguous elements. That is, if `batch == 16`, then axis `contig` will contain batches of 16 contiguous elements interleaved with axis `findfirst(isone.(striderank))`.
+ - `striderank` indicates the rank of the given stride with respect to the others. If for `A::T` we have `striderank[i] > striderank[j]`, then `stride(A,i) > stride(A,j)`.
```
ChrisRackauckas, line 127: add can_avx?

README.md
Tokazama: Do we also want to consider accommodating other memory layouts that have been used in improving performance of array operations [like these](https://github.com/JuliaMatrices/ArrayLayouts.jl/blob/master/src/memorylayout.jl#L54)?

chriselrod: @ChrisRackauckas 
"cpu" because I wanted it to be `false` for GPUArrays.
Given that [AbstractDeviceArrays <: DenseArray](https://github.com/JuliaGPU/GPUArrays.jl/blob/b34a724be5b4596ac74636adc01fac6768aa30e4/src/device/abstractarray.jl#L15) to opt into the `StrideArray` interface (even though `LinearAlgebra` routines presumably don't work), I wanted to make it clear they shouldn't opt into this.

@Tokazama 
The `stridelayout` function can efficiently handle row-major or increasing vs decreasing strides, e.g.:
```julia
julia> using ArrayInterface: stridelayout

julia> is_row_major(x) = stridelayout(x) === (2,1,(2,1))
 is_row_major (generic function with 1 method)

julia> B = rand(3,4);

julia> @code_typed is_row_major(B')
 CodeInfo(
1 ─     return true
) => Bool

julia> @code_typed is_row_major(B)
 CodeInfo(
1 ─     return false
) => Bool
```
But perhaps I should add density. Knowing that certain dims are dense could potentially allow subsets of loops to undergo `CartesianIndexing` -> `LinearIndexing` style loop fusion, even if on the whole the `IndexStyle` is `IndexCartesian()`.

Tokazama: Yeah, I was specifically thinking of dense arrays but didn't want to chance overlooking something else. Using IndexStyle to accomplish this is pretty clever.

chriselrod: I added density info on a per-stride basis.
```julia
julia> A = rand(3,4,5);

julia> stridelayout(PermutedDimsArray(A,(3,1,2)))
 (2, 1, (3, 1, 2), (true, true, true))

julia> stridelayout(@view(PermutedDimsArray(A,(3,1,2))[2,1:2,:]))
 (1, 1, (1, 2), (true, false))

julia> stridelayout(@view(PermutedDimsArray(A,(3,1,2))[2:3,1:2,:]))
 (2, 1, (3, 1, 2), (false, false, true))
```
Unfortunately, we can't tell from type info whether a slice of an array was partial or complete:
```julia
julia> s2 = Base.Slice(Base.OneTo(2))
 Base.Slice(Base.OneTo(2))

julia> view(A,s2,s2,s2)
 2×2×2 view(::Array{Float64,3}, :, :, :) with eltype Float64:
[:, :, 1] =
 0.329658  0.543774
 0.350255  0.0817058

[:, :, 2] =
 0.310211  0.126501
 0.587191  0.884039
```
as `Base.Slice{Base.OneTo{Int64}}` is the type of a `:` slice. This means we have to be very conservative with base `SubArray`s, but packages (like PaddedMatrices.jl) might want to provide more compile time info for their types.

The finer granularity would allow for the partial linear-vs-cartesian indexing.
Other reasons to have this information:
- using linear indexing to decide what to do for something like a `vec` implementation is a little goofy.
- LoopVectorization often spills scalar integer registers (i.e., `r*` registers). Many of these registers are used for holding array strides, and integer multiples of these strides. E.g., with a AVX512 column-major matmul kernel for `C = A * B`, it wants to hold `stride(A,2), 3stride(A,2), 5stride(A,2), stride(C,2), 3stride(C,2), 5stride(C,2)`. So if we know that `stride(A,2) == stride(C,2)`, we could free 3 registers, save a couple calculations, and a bunch of integer load/stores from the stack when we get less spills. I'm not sure how to implement this optimization / let a user guarantee that they aren't writing 100% valid code like
```julia
A = rand(4,5);
C = rand(40,50);
for m in axes(A,1), n in axes(A,2)
    C[m,n] = 2A[m,n]
end
```
So knowing that `A` and `C` are both dense still isn't enough to guarantee that the strides are equal.

Not particularly relevant to this issue, but I wonder if there'd be a  good way to communicate size equality. Users will often check for equal sizes in front of a loop, and `eachindex` will guarantee this as well
```julia
for I in eachindex(A,C) # checks for equivalent size
   
end
```
Because `eachindex` isn't always appropriate (e.g., when you don't have exact correspondence between axes of the arrays), I think it'd be a good to have something like an `eachindexalongaxis`:
```julia
function mymatmulkernel!(C, A, B)
    @avx for m in eachindexalongaxis((C,A),(1,1)), n in eachindexalongaxis((C,B),(2,2)), k in eachindexalongaxis((A,B),(2,1))
        C[m,n] += A[m,k] * B[k,n]
    end
end
```
where it checks sizes. `eachindexalongaxis` or `each_index_along_axis` doesn't really roll off the tongue, and I don't know where such a function should live.
I'd also have to think a little more about how to use it and actually pass the information of dimension equality (to be combined with density guarantees) to deduce stride-equality. The hacky way would be to just pattern match `eachindex` and `eachindexalongindex` in the macro, handling dimension equality through the expression, while handling density info based on types.

Tokazama: I'll make a separate issue about the axes specific indexing.

Tokazama: The only think I have left to say about this one is that you may want to include a some explanation or a link so people understand the "cpu" part. It's obviously not a huge deal, but I anticipate it becoming a common question that becomes irksome to continually explain.

chriselrod: How about a comment simply saying that it should return `false` for any sort of GPUArray?

Tokazama: That would work.

chriselrod:
This would probably be good to merge, but I am currently rewriting VectorizationBase and SIMDPirates (combining them into one library, just VectorizationBase) and the parts of LoopVectorization that will need to change to properly support this and a few other modifications.
I've also started rewriting PaddedMatrices, which will take advantage of these features and allow users to easily change the underlying memory layout.

They'll all support/make use of ArrayInterface.
I figured I'd leave the open until I'm closer to completion to give others the chance to comment, and myself the chance to realize that maybe something more/else is needed.

Once all that is done, I'll finish the fast convolution example.

Another change I'm working on (that contributes a lot to the overall code-churn) is representing unrolled vector operations with a type, rather than successive expressions. This will have a few advantages, e.g.
1. It'll interleave the operations within called functions. Many called functions, e.g. `tanh`, are composed of a long chain of operations. Instead of calling `tanh` 4 times, it'd allow for better out-of-order execution to interleave the individual operations of 4 calls, as would happen automatically with the unrolled type.
2. By applying operations group-wise, we can define specific overloads to optimize certain cases when possible. For example, replacing certain non-contiguous loads/stores with contiguous ones + sequences of shuffles.
To get full use out of this, it'd also sometimes start lumping all the distinct loads/stores from an array together in that bundled operation. That way something like this:
```julia
@avx for i in 1:length(x) >>> 1
    a = x[2i-1]
    b = x[2i]
    # do something with `a` and `b`
end
```
which currently results in 2 `gather`s, 1 each for `a` and `b`, can be optimized into 2 regular loads + 2 shuffles.

timholy:
Personally I'd be tempted to orthogonalize traits as much as possible. So something like  `Device(::Type) = CPU()` would mean it should be handled on the CPU, and then the striding would be a separate trait. There's a stab at this kind of interface in https://github.com/timholy/ComputationalResources.jl but it hasn't seen widespread usage. I'd be happy to have anything useful in that package migrate here if folks want.

Back to the orthogonality, for the trait depot you can do something like this::
```julia
myalgorithm(A::AbstractArray) = _myalgorithm((Device(A), StridedLayout(A), some_other_trait(A)), A)
_myalgorithm(::Tuple{CPU,Contiguous{1},<:Any}, A) = # implementation for is_cpu_columnmajor
_myalgorithm(::Tuple{CPU,Contiguous{2},<:Any}, A) = # implementation for row-major
...
```
or if you want to do some things using constant propagation of course there's the lovely tuple-destructuring syntax:
```julia
_myalgorithm((dev, layout, other)::Tuple{CPU,<:Any,<:Any}, A) = if layout === (1, (whatever)) ... else ... end
```
though it would be worth double-checking whether constant-propagation works flawlessly here.

src/ArrayInterface.jl
```diff
@@ -518,6 +518,146 @@ function restructure(x::Array,y)
   reshape(convert(Array,y),size(x)...)
 end
 
+"""
+is_cpu_column_major(::Type{T})
+
+Does an Array of type `T` point to column major memory in the cpu's address space?
+If `is_cpu_column_major(typeof(A))` return `true` and the element type is a primite
```
Tokazama, line 525: is `primite` type suppose to be `primitive`

src/ArrayInterface.jl
```diff
@@ -518,6 +518,146 @@ function restructure(x::Array,y)
...
+"""
+is_cpu_column_major(x) = is_cpu_column_major(typeof(x))
+is_cpu_column_major(::Type) = false
+is_cpu_column_major(::Type{<:Array}) = true
+is_cpu_column_major(::Type{S}) where {A, S <: SubArray{<:Any,<:Any,A,<:Tuple{Vararg{Union{Int,<:AbstractRange}}}}} = is_cpu_column_major(A)
+
+"""
+stridelayout(::Type{T}) -> (contig, batch, striderank, axesdense)
+
+Descrive the memory layout of a strided container of type `T`. If unknown or not strided, returns `nothing`.
```
Tokazama, line 541: `Descrive` -> `Describe`

Tokazama:
@timholy , Are you proposing that stride rank, batch, and contiguousness be separate traits or just device and stride layout?

timholy:
More the general principle that `this_and_that` traits are probably better split into their component parts. I acknowledge that some distinctions are fuzzy.

chriselrod:
I definitely agree with replacing `is_cpu_column_major` with `Device`.

I'll think more about `striderank`. Orthoganlization is good thing.

I'll try to start implementing some functions using this tonight.
Relying on constant prop makes me nervous, and I've been trying to abstain from generated functions as much as is reasonably possible (relying more on `@eval` in for loops, but I don't really know how this compares in precompilation time, package load time, or latency).

That does mean I'll likely want to make all of these types.
One thing I have been doing, for example, is define functions calculating linear indices from cartesian indices using the following pattern (e.g., for `getindex(A, i::Vararg)` call with `tdot(i, strides(A))` to take the dot product of the cartesian indices and array strides):
```julia
tdot(a::Tuple{A}, b::Tuple{B}) where {A,B} = a * b
tdot(a::Tuple, b::Tuple) = first(a) * first(b) + tdot(Base.tail(a), Base.tail(b))
```
with modifications for each `stridedpointer` type to avoid promoting the `_MM` type (used to represent vector loads) for contiguous dimensions into a vector of indices.
When the final load/store happens, a vector of indices will use a slow gather/scatter operation to load from/store to each element of that vector, while the `_MM` type results in fast contiguous loads/stores.

There were special `AbstractStridedPointer`s representing different memory layouts, e.g. one for column major (avoid promoting `_MM => Vec` for first dim only), another for transposed column major (avoid promoting for second), another for `SubArrays` without contiguous (always promote), etc.

So a goal is to replace all of these with a single `StridedPointer` type, parameterized by information from `ArrayInterface`. To add support, one then just has to define `Device` and `stridelayout` (or whatever `stridelayout` is broken up into), aside from the usual methods like `strides`.
Making these types would help
```julia
julia> struct Contiguous{N} end; Base.@pure Contiguous(N::Int) = Contiguous{N}.instance
 Contiguous

julia> totuple(::Contiguous{N}, ::Val{D}) where {N,D} = ntuple(d -> Val(d == N), Val(D))
 totuple (generic function with 1 method)

julia> @code_warntype totuple(Contiguous(2), Val(4))
Variables
  #self#::Core.Const(totuple, false)
  #unused#@_2::Core.Const(Contiguous{2}(), false)
  #unused#@_3::Core.Const(Val{4}(), false)
  #5::var"#5#6"{2}

Body::Tuple{Val{false},Val{true},Val{false},Val{false}}
1 ─ %1 = Core.apply_type(Main.:(var"#5#6"), $(Expr(:static_parameter, 1)))::Core.Const(var"#5#6"{2}, false)
│        (#5 = %new(%1))
│   %3 = #5::Core.Const(var"#5#6"{2}(), false)
│   %4 = Main.Val($(Expr(:static_parameter, 2)))::Core.Const(Val{4}(), false)
│   %5 = Main.ntuple(%3, %4)::Core.Const((Val{false}(), Val{true}(), Val{false}(), Val{false}()), false)
└──      return %5
```
Now I'd be able to do something like
```julia
tdot(a::Tuple{A}, b::Tuple{B}, ::Tuple{Val{false}}) where {A,B} = a * b
tdot(a::Tuple, b::Tuple, c::Tuple{Val{false},Vararg}) = first(a) * first(b) + tdot(Base.tail(a), Base.tail(b), Base.tail(c))
tdot(a::Tuple{A}, b::Tuple{B}, ::Tuple{Val{true}}) where {A,B} = mul_no_promote_MM(a, b)
tdot(a::Tuple, b::Tuple, c::Tuple{Val{true},Vararg}) = mul_no_promote_MM(first(a), first(b)) + tdot(Base.tail(a), Base.tail(b), Base.tail(c))
```
I may as well separate `Contiguous`, `BatchSize`, `StrideRank`, and `Dense` as well, depending on feedback and my thoughts as I start implementing these.

Tokazama:
I had some quick things I tested when developing the `indices` stuff to make sure it was flexible enough for what you're doing. I'll share some of it when I get to my computer.

chriselrod:
It's a little goofy that `Device` returns `nothing` for tuples or `SArray`s, so I'm open to a better name if there's one that would make this more clear.

What I'm going to use it for is just to know whether or not `pointer` works.

Tokazama:
Most of what follows was motivated by the comment you had under `stride_rank` and `dense_dims` where you said that there needs to be a better way of doing things, but it looks like it may be helpful in other places too. I wonder if what's missing is a simple way to map a dimension argument to the parent dimension and characterize the.

If we index something with `x` does it result in a single element (i.e. will that dimension be dropped)?
```julia
is_element(x) = is_element(typeof(x))
is_element(::Type{T}) where {T<:Number} = true
is_element(::Type{T}) where {T<:CartesianIndex} = true
is_element(::Type{T}) where {T} = false
@generated is_element(::Type{T}) where {T<:Tuple} = (map(is_element, I.parameters)...,)
```

Generic way to move to the parent dimensions space.
```julia
to_parent_dim(::Type{T}, ::Val{dim}) where {T,dim} = to_parent_dim(T, dim)
to_parent_dim(::Type{T}, dim::Integer) where {T<:AbstractArray} = dim
to_parent_dim(::Type{T}, dim::Integer) where {T<:Union{<:Adjoint,<:Transpose}} = (2, 1)[dim]
to_parent_dim(::Type{<:PermutedDimsArray{T,N,I}}, i::Integer) where {T,N,I} = I[i]
function to_parent_dim(::Type{<:SubArray{T,N,A,I}}, dim::Integer) where {T,N,A,I}
    # if a view is created with something that maps to a single value thena dimension
    # is dropped. We add all dimensions dropped up to i
    return i + accumulate(+, is_element(I))[dim]
end
```

Then `stride_rank` becomes something like
```julia
stride_rank(::Type{T}, ::Val{dim}) where {T,dim} = stride_rank(T, dim)
stride_rank(::Type{T}, dim::Integer) where {T} = nothing
stride_rank(::Type{T}, dim::Integer) where {T<:Array} = i
function stride_rank(::Type{T}, i::Integer) where {T<:AbstractArray}
    if parent_type(T) <: T  # assume no parent
        return nothing
    else
        return stride_rank(parent_type(T), to_parent_dim(T, dim))
    end
end
```

I'm not sure if this is what you had in mind  as an easier/better way for getting at those values, so if it's not helpful feel free to ignore it.

chriselrod:
That implementation (with a couple fixes below) is a bit more elegant
```julia
is_element(x) = is_element(typeof(x))
is_element(::Type{T}) where {T<:Number} = true
is_element(::Type{T}) where {T<:CartesianIndex} = true
is_element(::Type{T}) where {T} = false
@generated is_element(::Type{T}) where {T<:Tuple} = (map(is_element, T.parameters)...,)

to_parent_dim(::Type{T}, ::Val{dim}) where {T,dim} = to_parent_dim(T, dim)
to_parent_dim(::Type{T}, dim::Integer) where {T<:AbstractArray} = dim
to_parent_dim(::Type{T}, dim::Integer) where {T<:Union{<:Adjoint,<:Transpose}} = (2, 1)[dim]
to_parent_dim(::Type{<:PermutedDimsArray{T,N,I}}, i::Integer) where {T,N,I} = I[i]
function to_parent_dim(::Type{<:SubArray{T,N,A,I}}, dim::Integer) where {T,N,A,I}
    # if a view is created with something that maps to a single value thena dimension
    # is dropped. We add all dimensions dropped up to i
    return dim + accumulate(+, is_element(I))[dim]
end

stride_rank(::Type{T}, ::Val{dim}) where {T,dim} = stride_rank(T, dim)
stride_rank(::Type{T}, dim::Integer) where {T} = nothing
stride_rank(::Type{T}, dim::Integer) where {T<:Array} = dim
function stride_rank(::Type{T}, dim::Integer) where {T<:AbstractArray}
    if parent_type(T) <: T  # assume no parent
        return nothing
    else
        return stride_rank(parent_type(T), to_parent_dim(T, dim))
    end
end

stride_rank(::Type{T}) where {D, T <: AbstractArray{<:Any,D}} = ntuple(d -> stride_rank(T, d), Val{D}())
stride_rank(A) = stride_rank(typeof(A))
```
The difference in behavior is whether stride ranks should map to the original parents, or whether `sort(collect(stride_rank)) == 1:length(stride_rank)`:
```julia
julia> A = rand(2,3,4,5);

julia> stride_rank(@view(PermutedDimsArray(A, (3,2,4,1))[:,1,:,:]))
(3, 4, 1)

julia> ArrayInterface.stride_rank(@view(PermutedDimsArray(A, (3,2,4,1))[:,1,:,:]))
(2, 3, 1)

julia> stride_rank(@view(PermutedDimsArray(@view(PermutedDimsArray(A, (3,2,4,1))[:,1,:,:]), (2,3,1))[:,:,1])')
(1, 4)

julia> ArrayInterface.stride_rank(@view(PermutedDimsArray(@view(PermutedDimsArray(A, (3,2,4,1))[:,1,:,:]), (2,3,1))[:,:,1])')
(1, 2)
```
While the `(1, 4)` provides the same information + more, it may be less convenient to use in that it won't immediately give us the order of the ranks in the array view we have.
But maybe that's okay. I could write a function to produce `(1,2)` from `(1,4)`, but for some of the use cases I had in mind, the `(1,4)` is probably actually more useful. E.g., for ordering loops, the `(1,4)` let's us know that the stride across dim 2 is much larger than across dim 1, so it seems like it more faithfully represents the question it's answering...
Or what if we had `stride_rank(A) = (2,1,3)`, and `contiguous_axis(A) == Contiguous{1}()`.
This would mean that axis 1 has contiguous batches, but the stride between these batches is larger than the stride for axis 2.
Now if we take `B = @view(A[:,2,:])`, having `stride_rank(B) == (1,2)` with ``contiguous_axis(B) == Contiguous{1}()` would be misleading with respect to the actual layout. `contiguous_batch_size(B) !== ContiguousBatch{0}` would be a giveaway that this is the case, but still seems likely to result in bugs.

With all that in mind, I decided I like that behavior better and updated the PR.

Tokazama:
Do we need to have more device types or should that be left to other dedicated package? @timholy , you might have some thoughts on that. If we want to explore that more in a future PR instead that's fine, as long as the current implementation isn't prohibitive to necessary changes (e.g., ComputationalResources.jl has [`AbstractCPU`](https://github.com/timholy/ComputationalResources.jl/blob/master/src/ComputationalResources.jl#L35))

Tokazama:
Do we want to get this PR in before the next release or should I go ahead and make a PR for v2.12 now?

chriselrod:
It will probably be several weeks before I'm ready to release the VectorizationBase/LoopVectorization/PaddedMatrices taking advantage of this PR. I'd be inclined to release the 4 (4 = those 3 + ArrayInterface) together, as I may later on realize I want to make changes. And chose changes won't be breaking so long as the changed code wasn't merged.

In other words, let's go ahead and release now instead of waiting until those are ready.

If there's something from here you'd like to use, we can also merge that earlier.

chriselrod:
You may want to take a look at 
https://github.com/SciML/ArrayInterface.jl/pull/61/files#diff-0e0d0bdb9ecb9d7c302aa84bbe99cb37
It implements a statically known number (subtyping number), and I think overall it simplifies that static/dynamic tuple (i.e., replace that with a normal tuple that holds a mix of `Int` and `Static`). Adding it reduced the number of lines of code vs the `SDTuple`.

Being a number makes it easy to support arithmetic (it'll promote like an integer), and generally lets us avoid type piracy. E.g., we can define `:(:)(::Static, ::Static)` to make a  `OptionallyStaticUnitRange`.

I'll add tests soon.

Tokazama:
[StaticNumbers.jl](https://github.com/perrutquist/StaticNumbers.jl) does something similar. I think it makes sense to have a very basic "static" numbers interface here where we only implement the integer variant. Then we could allow other packages to expand on that as needed.

Unless we want to make this package responsible for centralizing the implementation of static ranges and numbers we should probably be careful not to put too much of that here. It's easy for that to get very involved very quickly. To be clear, I'm not against going that route. I just don't know if that's what everyone else has envisioned.

chriselrod:
I agree. I'm not sure at what point we should split it off into a separate package.
What I want generic functions for returning static size information in a simple to use way.

Making them tuples of numbers, and letting some of those numbers happen to be static is the best way of doing this I've thought of so far.
```julia
# Simple example implentation
struct MyArrayType{S,T,N} <: AbstractArray{T,N}
    data::Vector{T}
    sz::S
end
# Does the interface require an `NTuple{N,Int}`; must we use `ArrayInterface._get` the Static numbers?
Base.size(A::MyArrayType) = map(ArrayInterface._get, A.sz)
ArrayInterface.sdsize(A::MyArrayType) = A.sz

const MyVector{T} = MyArrayType{Tuple{<:Any},T,1}
const MyMatrix{T} = MyArrayType{Tuple{<:Any,<:Any},T,2}

const MyStaticVector{L,T} = MyArrayType{Tuple{Static{L}},T,1}
const MyStaticMatrix{M,N,T} = MyArrayType{Tuple{Static{M},Static{N}},T,2}
```
But this still has a disadvantage over the implementation with separate static and dynamic tuples, in that it's more awkward to refer to dynamic or static dimensions.
```julia
const MyInformativeMatrix{M,N,T} = MyArrayType{Tuple{M,N},T,2}

foo(A::MyInformativeMatrix{Static{M},Int}) where {M} = # first dim static, second dynamic
```
Right now, my plan is to define my array types following the `MyInformativeMatrix` approach.

Have any thoughts on the most ergonomic approach?

Tokazama:
I actually think that a lot of array packages could be simplified if we took a slightly more general approach like this:

```julia
struct ArrayType{T,N,Axs<:Tuple} <: AbstractArray{T,N}
    data::Vector{T}
    axes::Axs
end

const SOneTo{L} = OptionallyStaticUnitRange{Int,Val{1},Val{L}}
```

This allows `SOneTo` to basically be the same as `Static` where you can use `length` to unwrap the static number. The nice thing about this approach is that it allows you to also create offset/centered/padded indices without creating a new array type.

chriselrod:
The actual implementations will also carry stride information, and ranges aren't as good a conceptual fit there.

Tokazama:
> LLVM is bad at optimizing unrolled cartesian indexing code itself, so LoopVectorization has to emit a very particular kind of code to get good quality assembly, and this is reliant on extensive use of static indices.
Using static numbers for the indices is a separate issue from using static numbers for the strides. But given that the tuple_dot and related functions all must already be written to accommodate and optimize them, it'd be good represent static strides the same way. That just requires a general interface/support, which is why I put the code here.

I can see how strides makes ranges less than an ideal way of doing this. I think if we ensure that this is an internal implementation detail then there's no problem, but if we expect users to interact with this static number then we have to commit to implementing it here or in another package where it will inevitably become a lot more involved (e.g., someone wants to divide two static numbers and now we need to support static floats numbers). 

I can definitely see how representing it with size and strides is more directly applicable. 

> But I do recall the OffsetArray implementation involving a lot of tricky details to get it working correctly. I'd have to look into it more to see what it'd take to fold that in with our generic array representations.

It's not a huge deal for you to directly support. In AxisIndices I actually have replaced OffsetArrays.jl by using this trick. Once all this goes through hopefully I can build in some support for LoopVectorization.jl

chriselrod:
> if we expect users to interact with this static number then we have to commit to implementing it here or in another package where it will inevitably become a lot more involved (e.g., someone wants to divide two static numbers and now we need to support static floats numbers).

This is the current behavior of this PR:
```julia
julia> using ArrayInterface: Static

julia> Static(5) + Static(2)
 Static{7}()

julia> Static(5) ÷ Static(2)
 Static{2}()

julia> Static(5) / Static(2)
 2.5
```
which I think strikes a fair compromise, and lets code work as intended when you substitute random integers for their `Static` equivalents.
A full fledged static number implementation may of course prefer the floats to be static too so that it can implement optimized methods such as `Base.:(^)(::Static, ::Real)`.
But my optimizations are only related to indexing (primarily converting pointers from one type to another, to get LLVM to address memory in the way I want it to; memory is addressed with indexes in the form of `a * b + c` where `a` and `c` both must be constants, and `a` is only allowed to equal 0, 1, 2, 4, or 8; to get LLVM to choose the correct multiple, I need to cast pointers to a data type with the corresponding number of bytes, calculate the address, and then cast back).

But as I'm just using them for things related to indexing (including strides and now offsets), I don't need static-floats and promotion. I think most of the uses of static floats would probably be special function related, which seem out of scope for an Array Interface library.

> It's not a huge deal for you to directly support. In AxisIndices I actually have replaced OffsetArrays.jl by using this trick. 

Okay, great. I decided to go ahead and support offsets.

> Once all this goes through hopefully I can build in some support for LoopVectorization.jl

This PR intends to make that very easy to do.

src/stridelayout.jl
```diff
@@ -0,0 +1,260 @@
...
+permute(t::NTuple{N}, I::NTuple{N,Int}) where {N} = ntuple(n -> t[I[n]], Static{N}())
+@generated function permute(t::Tuple{Vararg{Any,N}}, ::Val{I}) where {N,I}
+    t = Expr(:tuple)
+    foreach(i -> push!(t.args, Expr(:ref, :t, i)), I)
+    Expr(:block, Expr(:meta, :inline), t)
+end
+
+sdsize(::Any) = nothing
+sdstrides(::Any) = nothing
+sdoffsets(::Any) = (Static{1}(),) # Assume arbitrary Julia data structures use 1-based indexing by default.
```
Tokazama, line 214: Offset indexing should have `indices` that are offset so you could probably just do

```julia
function sdoffsets(x, i)
    inds = indices(x, i)
    start = known_first(inds)
    if start === nothing
        return first(inds) - 1
    else
        return start - Static(1)
    end
end

sdoffsets(x) = ntuple(i -> sdoffsets(x, i), Val(ndims(x)))
```

This way any offset indexing is generically supported if `indices` is properly defined.

chriselrod: I do want this to correspond to "X-based indexing", so that a 1-based index Array has a value of "1".
This allows, [for example](https://github.com/chriselrod/VectorizationBase.jl/blob/316209a3bb761fa09106b143586a4f174990903c/src/strided_pointers/stridedpointers.jl#L103):
```julia
tdot(map(-, i, sdoffsets(ptr)), strides(ptr), contiguous_axis_indicator(ptr))
```
to be used for calculating the linear index, rather than needing to do this offsetting here.

I'll have to clarify in the documentation then that offset means "offset from 0" and not "offset from 1".

Additionally, `known_first` currently doesn't return a `Static`.
I would be in favor of making the `known_*` functions return `Static`s, although that change would be breaking.
Do you agree?

I just added:
```julia
@inline function sdoffsets(x, i)
    inds = indices(x, i)
    start = known_first(x)
    isnothing(start) ? first(inds) : Static(start)
end
sdoffsets(A::AbstractArray{<:Any,N}) where {N} = ntuple(n -> sdoffsets(A, n), Val{N}())
```
I'll add tests tonight.

Tokazama: > I'll have to clarify in the documentation then that offset means "offset from 0" and not "offset from 1".

Makes sense.

> I would be in favor of making the known_* functions return Statics, although that change would be breaking.

I'm on the fence about this one. Hopefully, it wouldn't be too breaking of a change since it should still be functionally equivalent.

Tokazama:
So we have `sdsize`, `sdstrides`, and `sdoffsets`. I understand this to mean that sizes could be static or dynamic and therefore of type `Int` or `Static`. Does this imply that `size`/`strides`/etc. as defined on new array types should only return `Int` unless explicitly having some sort `s` or `static` specifier in its name?

chriselrod:
> Does this imply that `size`/`strides`/etc. as defined on new array types should only return `Int` unless explicitly having some sort `s` or `static` specifier in its name?

Not necessarily. The primary reason for `sdsize` and `sdstrides` is to avoid committing type piracy.
I then thought it made sense to be consistent with `sdoffsets`.

You also  have to be careful when working with heterogeneously typed tuples to avoid type instabilities, so there is a risk that returning something other than `NTuple{N,Int}` for `size` causes someone else's code to have unexpectedly bad performance.
I'm not really sure how bad this risk is, but I'm also unsure of what the benefit of `Static` would be if someone's code wasn't written to deliberately take advantage of it; if a regular `Int` is known at compile time, that should normally be enough for Julia and LLVM to optimize it.

src/stridelayout.jl
```diff
@@ -0,0 +1,269 @@
...
+@generated function _dense_dims(::Type{S}, ::DenseDims{D}, ::StrideRank{R}) where {D,R,N,NP,T,A<:AbstractArray{T,NP},I,S <: SubArray{T,N,A,I}}
+    still_dense = true
+    sp = rank_to_sortperm(R)
+    densev = Vector{Bool}(undef, NP)
+    for np in 1:NP
+        spₙ = sp[np]
+        still_dense &= D[spₙ]
+        densev[spₙ] = still_dense
+        # a dim not being complete makes later dims not dense
+        still_dense && (still_dense = (I.parameters[spₙ] <: Base.Slice)::Bool)
```
mateuszbaran, line 190:
```suggestion
        still_dense &= (I.parameters[spₙ] <: Base.Slice)::Bool)
```
Should be the same and simpler?

src/stridelayout.jl
```diff
@@ -0,0 +1,269 @@
...
+@generated Base.sortperm(::StrideRank{R}) where {R} = rank_to_sortperm(R)
+
+stride_rank(x) = stride_rank(typeof(x))
+stride_rank(::Type) = nothing
+stride_rank(::Type{Array{T,N}}) where {T,N} = StrideRank{ntuple(identity, Val{N}())}()
+stride_rank(::Type{<:Tuple}) = StrideRank{(1,)}()
+
+stride_rank(::Type{B}) where {T, A, B <: Union{Transpose{T,A},Adjoint{T,A}}} = _stride_rank(B, stride_rank(A))
+_stride_rank(::Type{<:Union{Transpose{T,A},Adjoint{T,A}}}, ::Nothing) where {T,A<:AbstractMatrix{T}} = nothing
+_stride_rank(::Type{<:Union{Transpose{T,A},Adjoint{T,A}}}, rank) where {T,A<:AbstractMatrix{T}} = rank[Val{(2,1)}()]
```
mateuszbaran, line 128: Transposes and adjoints of `AbstractVector`s don't seem to be handled by `stride_rank`?

src/stridelayout.jl
```diff
@@ -0,0 +1,269 @@
...
+end
+
+permute(t::NTuple{N}, I::NTuple{N,Int}) where {N} = ntuple(n -> t[I[n]], Val{N}())
+@generated function permute(t::Tuple{Vararg{Any,N}}, ::Val{I}) where {N,I}
+    t = Expr(:tuple)
+    foreach(i -> push!(t.args, Expr(:ref, :t, i)), I)
+    Expr(:block, Expr(:meta, :inline), t)
+end
+
+sdsize(::Any) = nothing
```
mateuszbaran, line 212: Could you add a docstrings for `sdsize`, `sdstrides` and `sdoffsets`?

test/runtests.jl
```diff
@@ -191,6 +191,140 @@ end
     @test isone(ArrayInterface.known_step(typeof(1:4)))
 end
 
+@testset "Memory Layout" begin
```
mateuszbaran, line 194: Could you add some tests with adjoints and transposes?

chriselrod: There are a lot of tests with adjoints already.

src/ArrayInterface.jl
```diff
@@ -35,6 +35,8 @@ Otherwise, return `nothing`.
 known_length(x) = known_length(typeof(x))
 known_length(::Type{<:NTuple{N,<:Any}}) where {N} = N
 known_length(::Type{<:NamedTuple{L}}) where {L} = length(L)
+known_length(::Type{T}) where {T<:Base.Slice} = known_length(parent_type(T))
+known_length(::Type{<:Tuple{Vararg{Any,N}}}) where {N} = N
```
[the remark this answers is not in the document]
chriselrod: I'll remove the `NTuple` method.

test/runtests.jl
mateuszbaran: Hm, I don't see any in this testset.

src/ArrayInterface.jl
```diff
@@ -35,6 +35,8 @@ Otherwise, return `nothing`.
 known_length(x) = known_length(typeof(x))
 known_length(::Type{<:NTuple{N,<:Any}}) where {N} = N
 known_length(::Type{<:NamedTuple{L}}) where {L} = length(L)
+known_length(::Type{T}) where {T<:Base.Slice} = known_length(parent_type(T))
```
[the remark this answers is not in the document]
chriselrod: > I was thinking a bit about `Base.Slice` in the context of `HybridArrays` and my general conclusion was to treat is the same way as `Colon`. 

We are treating them as traversing the entire axis, so a `Base.Slice` preserves density.

> IIRC there was a case where a statically-known axis ended up being a slice with `Base.OneTo` instead of `SOneTo`

`SOneTo` is preferable, so I'd recommend getting that case to return a `Base.Slice(SOneTo(N))` instead.

> Are there any benefits of using this method instead?

Using which method instead of which alternative?

Currently, `known_length` takes a type and returns `nothing` (if length unknown, e.g. for `Base.Array`s), or returns the known length.
A problem is that we can't really use this within `@generated` functions, lest we face world age issues / defeat the entire purpose of providing an interface others can extend (I just checked and realize I have to fix `_sdsize`).

I'm considering a breaking change where `known_length` will return the `Static` numbers introduced in this PR, which will make it easier to "wrap" `@generated` functions (i.e., the wrapper can safely call `known_length` and friends, and with `Static`, it can pass the information on to the `@generated` function).
Although this works:
```julia
L = known_length(T)
if isnothing(L)
   # handle
else
  Static(L) # do something with this
end
```

src/stridelayout.jl
chriselrod: I'm thinking it should just return `(2,1)`?

src/ranges.jl
```diff
@@ -90,6 +90,10 @@ struct OptionallyStaticUnitRange{T,F,L} <: AbstractUnitRange{T}
   end
 end
 
+Base.:(:)(L, ::Static{U}) where {U} = OptionallyStaticUnitRange(L, Val(U))
```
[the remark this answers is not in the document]
chriselrod: Yes, I'll make that change and clean up the implementation a little.
I think it should be okay for `first` and `last` to also return `Static`, so that their definitions can be simplified to `r.start` and `r.stop`, respectively.
Then if `r` is an `OptionallyStaticUnitRange`, we should have that `r === first(r):last(r)`.

test/runtests.jl
chriselrod: Tests and line numbers featuring `'` within the "Memory Layout" testset:
```julia
200:    @test device(view(A, 1, :, 2:4)') === ArrayInterface.CPUPointer()
210:    @test @inferred(contiguous_axis(@view(PermutedDimsArray(A,(3,1,2))[2,1:2,:])')) === ArrayInterface.Contiguous(2)
213:    @test @inferred(contiguous_axis(@view(PermutedDimsArray(A,(3,1,2))[2:3,2,:])')) === ArrayInterface.Contiguous(-1)
214:    @test @inferred(contiguous_axis(@view(PermutedDimsArray(A,(3,1,2))[:,1:2,1])')) === ArrayInterface.Contiguous(1)
220:    @test @inferred(ArrayInterface.contiguous_axis_indicator(@view(PermutedDimsArray(A,(3,1,2))[2,1:2,:])')) === (Val(false),Val(true))
223:    @test @inferred(ArrayInterface.contiguous_axis_indicator(@view(PermutedDimsArray(A,(3,1,2))[2:3,2,:])')) === (Val(false),Val(false))
224:    @test @inferred(ArrayInterface.contiguous_axis_indicator(@view(PermutedDimsArray(A,(3,1,2))[:,1:2,1])')) === (Val(true),Val(false))
230:    @test @inferred(contiguous_batch_size(@view(PermutedDimsArray(A,(3,1,2))[2,1:2,:])')) === ArrayInterface.ContiguousBatch(0)
233:    @test @inferred(contiguous_batch_size(@view(PermutedDimsArray(A,(3,1,2))[2:3,2,:])')) === ArrayInterface.ContiguousBatch(-1)
234:    @test @inferred(contiguous_batch_size(@view(PermutedDimsArray(A,(3,1,2))[:,1:2,1])')) === ArrayInterface.ContiguousBatch(0)
240:    @test @inferred(stride_rank(@view(PermutedDimsArray(A,(3,1,2))[2,1:2,:])')) === ArrayInterface.StrideRank((2, 1))
243:    @test @inferred(stride_rank(@view(PermutedDimsArray(A,(3,1,2))[2:3,2,:])')) === ArrayInterface.StrideRank((2, 3))
244:    @test @inferred(stride_rank(@view(PermutedDimsArray(A,(3,1,2))[:,1:2,1])')) === ArrayInterface.StrideRank((1, 3))
250:    @test @inferred(dense_dims(@view(PermutedDimsArray(A,(3,1,2))[2,1:2,:])')) === ArrayInterface.DenseDims((false,true))
254:    @test @inferred(dense_dims(@view(PermutedDimsArray(A,(3,1,2))[2:3,2,:])')) === ArrayInterface.DenseDims((false,false))
255:    @test @inferred(dense_dims(@view(PermutedDimsArray(A,(3,1,2))[:,1:2,1])')) === ArrayInterface.DenseDims((true,false))
```

src/ArrayInterface.jl
```diff
@@ -565,6 +624,27 @@ function __init__()
     known_last(::Type{StaticArrays.SOneTo{N}}) where {N} = N
     known_length(::Type{StaticArrays.SOneTo{N}}) where {N} = N
 
+    device(::Type{<:StaticArrays.MArray}) = CPUPointer()
+    contiguous_axis(::Type{<:StaticArrays.StaticArray}) = Contiguous{1}()
```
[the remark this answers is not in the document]
chriselrod: Preferably it'd define a `parent` function, otherwise we'll need to access the parent through `A.data` ourselves, which may not be external API.

src/ArrayInterface.jl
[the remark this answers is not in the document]
mateuszbaran: > `SOneTo` is preferable, so I'd recommend getting that case to return a `Base.Slice(SOneTo(N))` instead.

OK, I'll keep that in mind.

> Using which method instead of which alternative?

In HybridArrays I take into account both array type and types of indices to figure out whether a certain access pattern is statically-sized or not. In some places for example `Colon` has to be special-cased, so the same special-casing can be used for `Slice`, like I did here for example: https://github.com/JuliaArrays/StaticArrays.jl/pull/783/files#diff-c1a2ec8ab9a030d018ed5124a63d38a4R193 . I don't see in what circumstances `known_length(::Type{<:Base.Slice})` would be used that don't treat `Colon` in a special way as well.

> I'm considering a breaking change where `known_length` will return the `Static` numbers introduced in this PR, which will make it easier to "wrap" `@generated` functions (i.e., the wrapper can safely call `known_length` and friends, and with `Static`, it can pass the information on to the `@generated` function).

I support this change. Generated functions that need static size information should get that information as an argument. That's what `StaticArray` does in many places. The downside main I think is potentially increasing compilation times.

src/ranges.jl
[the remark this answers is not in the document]
mateuszbaran: Yes, this makes sense.

src/stridelayout.jl
mateuszbaran: Yes, I think so as well.

test/runtests.jl
mateuszbaran: Right, I didn't look for apostrophes, sorry.

src/ArrayInterface.jl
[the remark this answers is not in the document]
Tokazama: If we don't have this defined for tuples then the method would return `nothing`, implying the length isn't known.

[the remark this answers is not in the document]
chriselrod: The `Tuple{Vararg{Any,N}}` definition works for both homogenous and heterogenous tuples, so the `NTuple`-specific definition was unnecessary.
```julia
julia> NTuple{1423} <: Tuple{Vararg{Any,1423}}
true
```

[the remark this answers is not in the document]
Tokazama: I misunderstood. I thought the final idea was to remove both `Tuple` methods, sorry.

src/ArrayInterface.jl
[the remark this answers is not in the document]
mateuszbaran: OK, I've added `parent` for `SizedArray` in my PR.

src/ArrayInterface.jl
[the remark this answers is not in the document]
mateuszbaran: To expand a little on Chris' answer, in case you didn't know:
```julia
julia> (1, 'd') isa NTuple{2,<:Any}
false

julia> (1, 'd') isa NTuple{2,Any}
true
```

Tokazama:
@chriselrod , this looks like it's maturing. Feel free to ping me once you're ready for a final review. At the risk of complicating things, it might be helpful to get some opinions from an interested developer from StaticArrays on the final review too.

chriselrod:
Might be convenient to get this merged so I can test dependent libaries on Travis and get test coverage.
So @Tokazama, feel free to review. 

> it might be helpful to get some opinions from an interested developer from StaticArrays on the final review too.

Have anyone in mind to ping?

Tokazama:
Perhaps @c42f would have some thoughts on this PR? I've seen him pop up a lot as I've lurked in the StaticArrays repository. I  think I remember him expressing interest in the use of static numbers and traits in some issue. Since the goal is largely to characterize dense linear arrays much of this could be helpful for indexing in StaticArrrays.

src/ArrayInterface.jl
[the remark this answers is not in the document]
mateuszbaran: Just a small reminder that you haven't fixed `_sdsize` yet: https://github.com/SciML/ArrayInterface.jl/pull/61/files#diff-b8b78add44fe2319a1282ed26bc78610R284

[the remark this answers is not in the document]
chriselrod: Oops, that slipped my mind. Added.

src/stridelayout.jl
```diff
@@ -0,0 +1,300 @@
...
+function rank_to_sortperm(R::NTuple{N,Int}) where {N}
+    sp = ntuple(zero, Val{N}())
+    r = ntuple(n -> sum(R[n] .≥ R), Val{N}())
+    @inbounds for n in 1:N
+        sp = Base.setindex(sp, n, r[n])
+    end
+    sp
+end
+"""
+rank_to_sortperm(::StrideRank) -> NTuple{N,Int}
```
Tokazama, line 115: Is this doc suppose to be for `sortperm` or for `rank_to_sortperm`?

src/stridelayout.jl
```diff
@@ -0,0 +1,300 @@
...
+
+Returns the `sortperm` of the stride ranks.
+"""
+@generated Base.sortperm(::StrideRank{R}) where {R} = rank_to_sortperm(R)
```
Tokazama, line 119: Is this docstring for `rank_to_sortperm` or for `sortperm(::StrideRank)`?

chriselrod: I moved it to `rank_to_sortperm`.

Tokazama:
LGTM

chriselrod:
> Edit: One thing I wanted to ask. Are we anticipating sdoffset being the official method for grabbing the offset of indices?

It is a kind of ugly name, but was chosen for consistency.
Should I break consistency and call it `offsets`?

If so, would there be better names for `sdsize` and `sdstrides`?
Perhaps it should be `ArrayInterface.size` and `ArrayInterface.strides`?

Tokazama:
> Should I break consistency and call it `offsets`?

I'd be fine with that.

> Perhaps it should be `ArrayInterface.size and `ArrayInterface.strides`?

I'd be fine with that. I hope that eventually we can just have `size(::StaticArray) -> Tuple{Vararg{<:Static}}` and methods that only accept `Int` will be generalized to accept `Static`. That way we don't have to specify all of those variations in here. I opened up [this issue](https://github.com/JuliaLang/julia/issues/37516) because ultimately we need a more formal interface for referring to dimensions in Base and it could allow us to pass `Static` to a methods like `axes` without redefining it for every array.

c42f:
> I hope that eventually we can just have size(::StaticArray) -> Tuple{Vararg{<:Static}} and methods that only accept Int will be generalized to accept Static.

Agreed this is the right way forward in the longer term. See also https://github.com/JuliaArrays/StaticArrays.jl/issues/807

Speaking of static integers, did you consider using `StaticNumbers.StaticInteger`?

chriselrod:
```sh
> juliamaster -O3 -e "@time using ArrayInterface"
  0.099558 seconds (126.48 k allocations: 8.313 MiB)

> juliamaster -O3 -e "@time using StaticNumbers"
  0.301221 seconds (635.21 k allocations: 35.458 MiB)

> julia -O3 -e "@time using ArrayInterface"
  0.177640 seconds (294.34 k allocations: 16.299 MiB, 2.82% gc time)

> julia -O3 -e "@time using StaticNumbers"
  0.177957 seconds (358.17 k allocations: 20.333 MiB, 2.80% gc time)
```
The load time seems relatively high.
ArrayInterface was on the current PR, which includes our simple `StaticNumber` implementation.

But I confess that my understanding of loading time is a little limited. Below, VectorizationBase is on a PR that depends on this ArrayInterface PR, using many methods:
```sh
> juliamaster -O3 -e "@time using VectorizationBase"
  0.028740 seconds (93.60 k allocations: 7.068 MiB)
```
0.03 seconds for `VectorizationBase` + `ArrayInterface`, but 0.1 seconds for `ArrayInterface` by itself?

But I wouldn't be opposed to us more broadly getting behind a package like `StaticNumbers`. If another project is using StaticNumbers, loading a half dozen different StaticNumber implementations would obviously be slower than loading just 1, besides the duplication of effort.

Tokazama:
Additionally, supporting other static numbers than `Int` was discussed. I think the conclusion was that it would make sense if it added in some meaningful way. However, the need for anything besides `Int` seems like a a pretty rare situation.

mateuszbaran:
That time added by `StaticNumbers` is largely caused by invalidations. After (breakingly) removing all sources of them I have:
```
➜  julia git:(master) ./julia -O3 -e "@time using StaticNumbers, ArrayInterface"
  0.260417 seconds (150.04 k allocations: 10.864 MiB)
➜  julia git:(master) ./julia -O3 -e "@time using ArrayInterface" 
  0.191187 seconds (63.91 k allocations: 4.931 MiB)
```
It doesn't look bad. Constructor of `Int64` from static numbers seems to be the biggest problem.

chriselrod:
There are a couple invalidations in ArrayInterface itself:
```julia
julia> using SnoopCompile

julia> invalidations = @snoopr using ArrayInterface;

julia> trees = invalidation_trees(invalidations)
2-element Vector{SnoopCompile.MethodInvalidations}:
 inserting promote_rule(::Type{Bool}, ::Type{S}) where S<:ArrayInterface.Static in ArrayInterface at /home/chriselrod/.julia/dev/ArrayInterface/src/static.jl:34 invalidated:
   backedges: 1: superseding promote_rule(::Type{Bool}, ::Type{T}) where T<:Number in Base at bool.jl:4 with MethodInstance for promote_rule(::Type{Bool}, ::Type{var"#s431"} where var"#s431"<:Integer) (2 children)

 inserting convert(::Type{T}, ::ArrayInterface.Static{N}) where {T<:Number, N} in ArrayInterface at /home/chriselrod/.julia/dev/ArrayInterface/src/static.jl:18 invalidated:
   backedges: 1: superseding convert(::Type{T}, x::Number) where T<:Number in Base at number.jl:7 with MethodInstance for convert(::Type{Int8}, ::Integer) (1 children)
              2: superseding convert(::Type{T}, x::Number) where T<:Number in Base at number.jl:7 with MethodInstance for convert(::Type{UInt8}, ::Integer) (1 children)
              3: superseding convert(::Type{T}, x::Number) where T<:Number in Base at number.jl:7 with MethodInstance for convert(::Type{Int64}, ::Integer) (9 children)
              4: superseding convert(::Type{T}, x::Number) where T<:Number in Base at number.jl:7 with MethodInstance for convert(::Type{UInt64}, ::Integer) (27 children)
```
Can we do anything about them?

chriselrod:
I'd like to set the goal of merging this on Friday.

I'm open to `StaticNumbers`. While all we need are `Static` integers, it's less code to `using StaticNumbers: StaticIntegers`.

Tokazama:
I'm not completely against StaticNumbers.jl (I've even brought it up in the past), but I'm not 100% sold on it. Even if we decide the niche use case for other static numbers is sufficient to justify supporting them, I'm not sure we're ready to handle things here like `StaticNumber:StaticNumber:StaticNumber`, which are a lot more complicated to properly support.

If we do decide to support all static numbers I think that's going to have to be a separate, large PR. Once you support different types of numbers for ranges that have different step sizes it gets pretty complicated (e.g., safely handling code with smaller byte sizes, handling high precision number types, accounting for numbers with units attached, etc.).

It seems like the most straightforward approach is to just use `Static` for now and then if it becomes clear that we need more static number support we can make the change. If methods like `ArrayInterface.size` and `ArrayInterface.static_first` are used as intended (generic support for static `Int`), then most other packages should be agnostic to the change. If we're really worried about it we could even just change `Static` to `StaticInteger` so it's a matter of `using StaticNumbers: StaticInteger` if the time should ever come we want to depend on StaticNumbers.

c42f:
> It seems like the most straightforward approach is to just use Static for now and then if it becomes clear that we need more static number support we can make the change.

FWIW I think it's completely fine to use `Static` for now, and I hesitate to depend on `StaticNumbers` in `StaticArrays` for similar reasons:
* We really only need a small subset of stable core functionality — `StaticInteger`, `static()` and perhaps a couple of other things.
* Load time
* `StaticNumbers` currently extends `StaticArrays` via `Requires.jl`. So disentangling that would be a little work.

Perhaps we should make a StaticNumbersBase.jl to define only the core functionality. @perrutquist what are your thoughts?

perrutquist:
I think a StaticNumbersBase package would be a good idea. I've been thinking for a while that I need to split up that package. (It started out as a generalization of Zeros.jl, and hence with real numbers, but the integer use-case is fairly orthogonal to that.)

Maybe even better if we could have a `StaticBase` which also contains stubs and abstract types for static arrays and iterators? That way it would be easier to define interaction without using Requires.

However, it seems that most of the load time (and post-load slowdown) in StaticNumbers.jl comes from a single method definition: `Int(::StaticInteger)`. Either there is a way to fix that, or load-times will be a problem for any implementation of static integers that can be used to construct an `Int`. cf: https://github.com/perrutquist/StaticNumbers.jl/issues/11

chriselrod:
I don't really have any experience here, but this by itself at least doesn't seem to take very long on Julia 1.6:
```julia
julia> using SnoopCompile

julia> invalidations = @snoopr begin
       struct Static{N} <: Integer end
       (::Type{T})(::Static{N}) where {T <: Number, N} = T(N)
       end;

julia> invalidation_trees(invalidations)
1-element Vector{SnoopCompile.MethodInvalidations}:
 inserting (::Type{T})(::Static{N}) where {T<:Number, N} in Main at REPL[2]:3 invalidated:
   mt_backedges:  1: signature Tuple{Type{Int64}, Integer} triggered MethodInstance for findnext(::Pkg.Types.var"#22#23"{String}, ::AbstractString, ::Integer) (0 children)
                  2: signature Tuple{Type{Int64}, Integer} triggered MethodInstance for _array_for(::Type{T}, ::Any, ::Base.HasLength) where T (0 children)
                  3: signature Tuple{Type{Int64}, Integer} triggered MethodInstance for to_shape(::Integer) (0 children)
                  4: signature Tuple{Type{Int64}, Integer} triggered MethodInstance for _array_for(::DataType, ::Any, ::Base.HasLength) (0 children)
                  5: signature Tuple{Type{Int64}, Integer} triggered MethodInstance for _similar_for(::Vector{_A} where _A, ::Type, ::Base.Generator{_A, Base.var"#811#813"
} where _A, ::Base.HasLength) (0 children)
                  6: signature Tuple{Type{Int64}, Integer} triggered MethodInstance for _similar_for(::Vector{_A} where _A, ::DataType, ::Base.Generator{_A, Base.var"#811#
813"} where _A, ::Base.HasLength) (0 children)
                  7: signature Tuple{Type{Int64}, Integer} triggered MethodInstance for _array_for(::Type{Base.RefValue{Any}}, ::Any, ::Base.HasLength) (0 children)
                  8: signature Tuple{Type{UInt16}, Integer} triggered MethodInstance for connect!(::Sockets.TCPSocket, ::Sockets.IPv4, ::Integer) (0 children)
                  9: signature Tuple{Type{UInt16}, Integer} triggered MethodInstance for connect!(::Sockets.TCPSocket, ::Sockets.IPv6, ::Integer) (0 children)
                 10: signature Tuple{Type{UInt32}, Integer} triggered MethodInstance for VersionNumber(::Integer, ::Int64, ::Int64, ::Tuple{}, ::Tuple{}) (1 children)
                 11: signature Tuple{Type{Int64}, Integer} triggered MethodInstance for manage(::Distributed.SSHManager, ::Integer, ::Distributed.WorkerConfig, ::Symbol) (
1 children)
                 12: signature Tuple{Type{UInt64}, Integer} triggered MethodInstance for convert(::Type{UInt64}, ::Integer) (7 children)
                 13: signature Tuple{Type{Int64}, Integer} triggered MethodInstance for prevind(::String, ::Integer) (11 children)
                 14: signature Tuple{Type{Int64}, Integer} triggered MethodInstance for full_va_len(::Core.SimpleVector) (784 children)
```
It seems odd to me that
```julia
julia> struct Static{N} <: Integer end

julia> @which Int(Static{5}())
ERROR: no unique matching method found for the specified argument types
```
there are no methods, yet we can't insert one without invalidating others?

I'd be in favor of a `StaticBase`.


While I like the brevity of `Static`, but could rename it `StaticInteger` so that we could swap the underlying implementation with a dependency on another package without breaking dependents.
(While `Static` is already on ArrayInterface master, there haven't yet been any releases featuring it.)

perrutquist:
I'm only speculating, but I think the invalidations come from methods that were defined when `Int` was only defined for a small number of `Integer` types, so methods that called `Int(x::Integer)` could be compiled with the assumption that `x` belonged to that small type union.
I also think that most of the slowdown does not happen immediately when the constructor is defined, but rather in the code that runs after it, as methods are recompiled.

If you're thinking of renaming `Static`, then maybe a good name would be `StaticSigned` or `StaticInt` to signify that the argument is always a `Signed` or an `Int`. In retrospect, it was a mistake to create `StaticInteger{X}`, since it is impossible to dispatch on the type of `X`, and this is causing me a big headache in trying to deal with unsigned integers. (In a future release, `StaticInteger` will likely be a type union.)

Tokazama:
I played with static numbers pretty extensively a while ago. There are all sorts of quirky things that will get you if you're not careful. I recall getting tripped up on ensuring type stability with a certain type of log method.

README.md
```diff
@@ -134,6 +134,74 @@ Otherwise, returns `nothing`. For example, `known_step(UnitRange{Int})` returns
...
+
+## stride_rank(::Type{T})
+
+Returns the rank of each stride.
+
+## dense_dims(::Type{T})
+Returns a tuple of indicators for whether each axis is dense.
+An axis `i` of array `A` is dense if `stride(A, i) * size(A, i) == stride(A, j)` where `j` is the axis (if it exists) such that `stride_rank(A)[i] + 1 == stride_rank(A)[j]`.
+
+## sdsize(A)
```
Tokazama, line 170: Change to ArrayInterface.size(A)?

README.md
```diff
@@ -134,6 +134,74 @@ Otherwise, returns `nothing`. For example, `known_step(UnitRange{Int})` returns
...
+```julia
+julia> using StaticArrays; using ArrayInterface: sdsize
+
+julia> A = @SMatrix rand(3,4);
+
+julia> sdsize(A)
+(Static{3}(), Static{4}())
+```
+
+## sdstrides(A)
```
Tokazama, line 183: Change to `ArrayInterface.strides(A)`?

README.md
```diff
@@ -134,6 +134,74 @@ Otherwise, returns `nothing`. For example, `known_step(UnitRange{Int})` returns
...
+these should be returned as `Static` numbers. For example:
+```julia
+julia> using ArrayInterface: sdstrides
+
+julia> A = rand(3,4);
+
+julia> sdstrides(A)
+(Static{1}(), 3)
+```
+## sdoffsets(A)
```
Tokazama, line 195: `offsets(A)`

chriselrod:
You have to be careful with `StaticInt`s and heterogeneous tuples in general, which `ArrayInterface.size` and `ArrayInterface.static` are likely to produce.

Tokazama:
> You have to be careful with StaticInts and heterogeneously-typed tuples in general, which ArrayInterface.size and ArrayInterface.static are likely to produce.

I'm assuming you mean the difference between `NTuple{N,<:Integer}` and `Tuple{Vararg{<:Integer,N}}`?

README.md
```diff
@@ -134,12 +134,80 @@ Otherwise, returns `nothing`. For example, `known_step(UnitRange{Int})` returns
...
+julia> ArrayInterface.size(A)
+(StaticInt{3}(), StaticInt{4}())
+```
+
+## ArrayInterface.strides(A)
+
+Returns the strides of array `A`. If any strides are known at compile time,
+these should be returned as `StaticInt`s. For example:
+```julia
+julia> using ArrayInterface: sdstrides
```
Tokazama, line 188: `strides`

README.md
```diff
@@ -134,12 +134,80 @@ Otherwise, returns `nothing`. For example, `known_step(UnitRange{Int})` returns
...
+julia> A = rand(3,4);
+
+julia> ArrayInterface.strides(A)
+(StaticInt{1}(), 3)
+```
+## offsets(A)
+
+Returns offsets of indices with respect to 0. If values are known at compile time,
+it should return them as `StaticInt`s.
+For example, if `A isa Base.Matrix`, `sdoffsets(A) === (StaticInt(1), StaticInt(1))`.
```
Tokazama, line 199: `offsets(A)`

README.md
```diff
@@ -134,12 +134,80 @@ Otherwise, returns `nothing`. For example, `known_step(UnitRange{Int})` returns
...
+## dense_dims(::Type{T})
+Returns a tuple of indicators for whether each axis is dense.
+An axis `i` of array `A` is dense if `stride(A, i) * size(A, i) == stride(A, j)` where `j` is the axis (if it exists) such that `stride_rank(A)[i] + 1 == stride_rank(A)[j]`.
+
+## ArrayInterface.size(A)
+
+Returns the size of `A`. If the size of any axes are known at compile time,
+these should be returned as `StaticInt`s. For example:
+```julia
+julia> using StaticArrays; using ArrayInterface: sdsize
```
Tokazama, line 175: `size`

chriselrod:
I meant
```julia
julia> using StaticArrays, ArrayInterface

julia> A = @SMatrix rand(4,3);

julia> @code_warntype getindex(ArrayInterface.size(A), 1)
Variables
  #self#::Core.Compiler.Const(getindex, false)
  t::Core.Compiler.Const((ArrayInterface.StaticInt{4}(), ArrayInterface.StaticInt{3}()), false)
  i::Int64

Body::Union{ArrayInterface.StaticInt{4}, ArrayInterface.StaticInt{3}}
1 ─      nothing
│   %2 = Base.getfield(t, i, $(Expr(:boundscheck)))::Union{ArrayInterface.StaticInt{4}, ArrayInterface.StaticInt{3}}
└──      return %2

julia> @code_warntype getindex(ArrayInterface.size(A), ArrayInterface.StaticInt(1))
Variables
  #self#::Core.Compiler.Const(getindex, false)
  t::Core.Compiler.Const((ArrayInterface.StaticInt{4}(), ArrayInterface.StaticInt{3}()), false)
  i::Core.Compiler.Const(ArrayInterface.StaticInt{1}(), false)

Body::ArrayInterface.StaticInt{4}
1 ─      nothing
│   %2 = Base.convert(Base.Int, i)::Core.Compiler.Const(1, false)
│   %3 = Base.getfield(t, %2, $(Expr(:boundscheck)))::Core.Compiler.Const(ArrayInterface.StaticInt{4}(), false)
└──      return %3
```

Issue #62, opened 2020-08-12
Safe co-iteration along axes


Tokazama:
This came up with the following comment in from #61 .

> Users will often check for equal sizes in front of a loop, and eachindex will guarantee this as well
> ```julia
> for I in eachindex(A,C) # checks for equivalent size
>    
> end
> ```
> Because eachindex isn't always appropriate (e.g., when you don't have exact correspondence between axes of the arrays), I think it'd be a good to have something like an eachindexalongaxis:
> 
> ```julia
> function mymatmulkernel!(C, A, B)
>     @avx for m in eachindexalongaxis((C,A),(1,1)), n in eachindexalongaxis((C,B),(2,2)), k in eachindexalongaxis((A,B),(2,1))
>         C[m,n] += A[m,k] * B[k,n]
>     end
> end
> ```
> where it checks sizes. eachindexalongaxis or each_index_along_axis doesn't really roll off > the tongue, and I don't know where such a function should live.
> I'd also have to think a little more about how to use it and actually pass the information of > dimension equality (to be combined with density guarantees) to deduce stride-equality. The hacky way would be to just pattern match eachindex and eachindexalongindex in the macro, handling dimension equality through the expression, while handling density info based on types.

I'd like to propose the use of `indices` for specifically retrieving the iterator that corresponds to each axis of an array. It would function like `axes` but if there's any sort of other data associated with an axis it throws it away so there's a clear relationship between memory layout and and each dimension.

In practice it would look like this:
```julia
julia> indices(ones(2,3))
(Base.OneTo(2), Base.OneTo(3))

julia> indices(ones(2,3), 1)
Base.OneTo(2)

julia> indices(ones(2,3), 2)
Base.OneTo(3)

julia> indices(ones(2,3), ones(2,3), 2)
Base.OneTo(3)

julia> indices(ones(2,3), ones(2,2), 2)
ERROR: DimensionMismatch("all inputs to indices must have the same indices, got Base.OneTo(3) and Base.OneTo(2)")
```

In this case the first several examples are the same as if using `axes`, but it helps for cases like those in AxisIndices.jl where each axis can have additional information (e.g., keys) associated with the indices. The last two examples would ensure that the iterator for each collection is the same along the second dimension.

chriselrod:
In my example I used a tuple of indices, so that we could
```julia
julia> indices(ones(2,3), ones(2,2), (1,1))
Base.OneTo(2)

[...]

PR #63, opened 2020-08-15
Safe co-iteration across an axis for 1+ arrays

Tokazama:
This should help with #62. The main idea is that we can get the convenience of checking compatible indices like in `eachindex(A...)`, but for specific axes. Possible concerns may be:

1. I decided to have `indices(x)` always return iterable indices instead of a tuple of indices. So the default behavior is like `eachindex`. This does two things for us:
    1. The behavior of `for i in indices(args...) ... end` is more predictable becaues `i` can always point to a single value of the array(s).
    2. It makes it very easy to incorporate/extend for other array/array-like types. If the behavior is unique to each axis then only `indices(x)` needs to be specified. If the behavior is unique for the array then only `indices(x, i)` needs to be specified.
2. I didn't implement any special behavior for combining/optimizing known range first/last/step although it was mentioned in #62. I figured even a minimal range type would be meaty enough for an independent PR. 

If people don't find the first point compelling enough to depart from behavior like `axes` I can change it. I just thought it was a really interesting idea. I'm also happy to implement a range if it's seen as immediately necessary for this PR.

### DIFF
--- a/src/ranges.jl
+++ b/src/ranges.jl
@@ -58,25 +58,25 @@ from other valid indices. Therefore, users should not expect the same checks are
 to ensure construction of a valid `OptionallyStaticUnitRange` as a `UnitRange`.
 """
 struct OptionallyStaticUnitRange{T,F,L} <: AbstractUnitRange{T}
-    start::F
-    stop::L
+  start::F
+  stop::L
 
-    function OptionallyStaticUnitRange{T}(start, stop) where {T<:Real}
-        if _get(start) isa T
-            if _get(stop) isa T
-                return new{T,typeof(start),typeof(stop)}(start, stop)
-            else
-                return OptionallyStaticUnitRange{T}(start, _convert(T, stop))
-            end
-        else
-            return OptionallyStaticUnitRange{T}(_convert(T, start), stop)
-        end
+  function OptionallyStaticUnitRange{T}(start, stop) where {T<:Real}
+    if _get(start) isa T
+      if _get(stop) isa T
+        return new{T,typeof(start),typeof(stop)}(start, stop)
+      else
+        return OptionallyStaticUnitRange{T}(start, _convert(T, stop))
+      end
+    else
+      return OptionallyStaticUnitRange{T}(_convert(T, start), stop)
     end
+  end
 
-    function OptionallyStaticUnitRange(start, stop)
-        T = promote_type(typeof(_get(start)), typeof(_get(stop)))
-        return OptionallyStaticUnitRange{T}(start, stop)
-    end
+  function OptionallyStaticUnitRange(start, stop)
+    T = promote_type(typeof(_get(start)), typeof(_get(stop)))
+    return OptionallyStaticUnitRange{T}(start, stop)
+  end
 end
 
 Base.first(r::OptionallyStaticUnitRange{<:Any,Val{F}}) where {F} = F
@@ -92,11 +92,11 @@ known_step(::Type{<:OptionallyStaticUnitRange{T}}) where {T} = one(T)
 known_last(::Type{<:OptionallyStaticUnitRange{<:Any,<:Any,Val{L}}}) where {L} = L
 
 function Base.isempty(r::OptionallyStaticUnitRange)
-    if known_first(r) === oneunit(eltype(r))
-        return unsafe_isempty_one_to(last(r))
-    else
-        return unsafe_isempty_unit_range(first(r), last(r))
-    end
+  if known_first(r) === oneunit(eltype(r))
+    return unsafe_isempty_one_to(last(r))
+  else
+    return unsafe_isempty_unit_range(first(r), last(r))
+  end
 end
 
 unsafe_isempty_one_to(lst) = lst <= zero(lst)
@@ -108,26 +108,26 @@ unsafe_length_one_to(lst::T) where {T<:Int} = T(lst)
 unsafe_length_one_to(lst::T) where {T} = Integer(lst - zero(lst))
 
 Base.@propagate_inbounds function Base.getindex(r::OptionallyStaticUnitRange, i::Integer)
-    if known_first(r) === oneunit(r)
-        return get_index_one_to(r, i)
-    else
-        return get_index_unit_range(r, i)
-    end
+  if known_first(r) === oneunit(r)
+    return get_index_one_to(r, i)
+  else
+    return get_index_unit_range(r, i)
+  end
 end
 
 @inline function get_index_one_to(r, i)
-    @boundscheck if ((i > 0) & (i <= last(r)))
-        throw(BoundsError(r, i))
-    end
-    return convert(eltype(r), i)
+  @boundscheck if ((i > 0) & (i <= last(r)))
+    throw(BoundsError(r, i))
+  end
+  return convert(eltype(r), i)
 end
 
 @inline function get_index_unit_range(r, i)
-    val = first(r) + (i - 1)
-    @boundscheck if i > 0 && val <= last(r) && val >= first(r)
-        throw(BoundsError(r, i))
-    end
-    return convert(eltype(r), val)
+  val = first(r) + (i - 1)
+  @boundscheck if i > 0 && val <= last(r) && val >= first(r)
+    throw(BoundsError(r, i))
+  end
+  return convert(eltype(r), val)
 end
 
 _try_static(x, y) = Val(x)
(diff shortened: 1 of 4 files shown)


chriselrod:
> I didn't implement any special behavior for combining/optimizing known range first/last/step although it was mentioned in #62. I figured even a minimal range type would be meaty enough for an independent PR.

I would like to see this.
I'd like to add that instead of the pattern matching, I'd like the iterators to also track an is-dense along the lines of https://github.com/SciML/ArrayInterface.jl/pull/61.

Then that would provide all the information I'd need to implement the optimizations I talked about, i.e.:
1. Possibility of flattening some subset of loops, CartesianIndexing -> LinearIndexing style.
2. Realizing that stride values must necessarily be equal between arrays, to avoid spilling integer registers.

Tokazama:
> I'd like to add that instead of the pattern matching, I'd like the iterators to also track an is-dense along the lines of #61.

Are you referring to something that does something like `stridelayout(::Type{T}) -> (contig, batch, striderank)` for each axis, or just flattening out the results of `indices` when possible?

chriselrod:
I forgot to add `axesdense` to the list of returns: `stridelayout(::Type{T}) -> (contig, batch, striderank, axesdense)`.
So that'd mean consolidating the `axesdense` information.
The reason I want that info from the axis iterator rather than to read it directly from the arrays is:
```julia
A = rand(10, 5);
for i in 1:3, j in 1:4
    A[i,j] *= 2
end
```
`A` is dense, but that doesn't mean these loops could be flattened.

Tokazama:
Are you thinking of something like this?

```julia
is_dense(::Type{T}, i::Int) where {T} = nothing
is_dense(::Type{<:Array}, i::Int) = true
@inline function is_dense(::Type{<:Union{Transpose{T,A},Adjoint{T,A}}}) where {T,A<:AbstractMatrix{T}}
    return is_dense(A, (2, 1)[i])
end
@inline function is_dense(::Type{<:PermutedDimsArray{T,N,I1,I2,A}}, i::Int) where {T,N,I1,I2,A<:AbstractArray{T,N}}
    return is_dense(A, I1[i])
end
```

I don't really see how an iterator along each dimension in isolation can incorporate the density since it requires knowing about the previous dimensions and strides.

chriselrod:
Something like:
```julia
is_dense(::Type{T}, i) where {T} = last(stridelayout(T))[i]
```

> I don't really see how an iterator along each dimension in isolation can incorporate the density since it requires knowing about the previous dimensions and strides.

You're right. Given what I said:
```julia
A = rand(2,3,4);
B = PermutedDimsArray(rand(4,3,2), (3,2,1));
for k in indices((A,B),3), j in indices((A,B),2), i in indices((A,B),1)
    # do something with A[i,j,k] and B[i,j,k]
end
````
These each of these should be dense, even though we obviously can't flatten the loops.
However, we could flatten them if we only had `A` or if we only had `B`.

That is, `is_dense` by itself does not provide enough information, but I think it does in conjunction with `stridelayout`.
I.e., in querying whether we can flatten two loops, we need to check:
1. That they share the same order among all arrays indexed by them (provided by `stridelayout`)
2. That the loop with the smaller stride `is_dense`/covers the full axis for each of them (information must be provided by the stride-iterator).

**Comment by Tokazama** [comment]
This sounds a lot like what `Slice` is for. Would it make sense to wrap whatever we get from `indices` into something like `Base.Slice` or should we have a unique type that's "co-slicing"?

**Comment by chriselrod** [comment]
Okay, if that's the understood meaning of `Slice`.

Simply doing that, and then checking `stridelayout` to see if all the sliced dimensions are dense sounds like a good approach to me.
Meaning `indices` only has to accumulate statically known `first`/`last`/`step`.

File: src/ranges.jl [hunk 1]
```diff
@@ -0,0 +1,139 @@
...
+  stop::L
+
+  function OptionallyStaticRange(start::F, step::S, stop::L) where {F,S,L}
+      T = promote_type(_eltype(F), _eltype(S), eltype(L))
+      return new{T,F,S,L}(start, step, stop)
+  end
+end
+
+Base.first(r::OptionallyStaticRange{<:Any,Val{F}}) where {F} = F
+Base.first(r::OptionallyStaticRange{<:Any,<:Any}) = getfield(r, :start)
```
chriselrod, line 56: I'm curious, what is the reason to prefer `getfield(r, :start)` over `r.start`?
I've always used the latter in my own code.

src/ranges.jl
```diff
@@ -0,0 +1,139 @@
...
+        return zero(T)
+    else
+        if known_step(r) === oneunit(T)
+            if known_first(r) === oneunit(T)
+                return last(r)
+            else
+                return last(r) - first(r) + step(r)
+            end
+        else
+            return Integer(div((last(r) - first(r)) + step(r), step(r)))
```
chriselrod, line 83: Maybe `Base.udiv_int` if we can guarantee the checks will pass?

src/ranges.jl
Tokazama: Sorry, it's just a habit from working with types that have overloaded the `getproperty` method for metadata like tables. I can change this, as it's probably easier to read `r.start`.

Tokazama:
One last question and I think I can get something more polished together. `Base.Slice` only accepts `AbstractUnitRange`. Do you need `OptinallyStaticRange` to be capable of having different step sizes?

src/ranges.jl
Tokazama: The code for calculating the length on `StepRange` is what I'm basing this one and I should probably expand this out to also include [this](https://github.com/JuliaLang/julia/blob/master/base/range.jl#L554) sort of code.

chriselrod:
Maybe add
```julia
struct OptionallyStaticUnitRange{T,F,L} <: AbstractUnitRange{T}
    start::F
    stop::L
end

maybe_slice(r) = r
function maybe_slice(r::OptionallyStaticRange{T,F,::Val{1},L}) where {T,F,L}
    Base.Slice(OptionallyStaticUnitRange{T,F,L}(first(r), last(r)))
end
```
Then you could use `maybe_slice(reduce(_pick_range, inds))`.

Tokazama:
I would really love if we had a common interface where someone could do this with a new type

```julia
function Base.length(r::NewRangeType{T}) where where{T}
    if isempty(r)
        return zero(T)
    else
        return ArrayInterface.unsafe_length(first(r), step(r), last(r))
    end
end
```

But that turns into a lot of code very quickly and I should probably save that for a different PR. Would it be okay if I just do a unit range thing for now and then we can see how everyone feels about a more aggressive range focused PR later that could include `OptionallyStaticStepRange`?

chriselrod:
> I would really love if we had a common interface where someone could do this with a new type

Do you mean, define an interface such that `length` will work in that manner?
Defining your own `abstract type` that they could subtype (along with defining `first`, etc) would be easiest, otherwise wouldn't you need a PR on base Julia to do this with traits without committing type piracy? Or am I missing something?

The `OptionallyStaticStepRange` is fine.
Are we going to make it so that the type promises to be a full slice? Or do you want to still wrap the returns in `Base.Slice`?

Tokazama:
> Do you mean, define an interface such that length will work in that manner?
Defining your own abstract type that they could subtype (along with defining first, etc) would be easiest, otherwise wouldn't you need a PR on base Julia to do this with traits without committing type piracy? Or am I missing something?

I think it would ultimately need to be in base but there are a lot of little problems with how ranges are implemented in base and we'd need to start by getting the `known_*` methods there first.

> Are we going to make it so that the type promises to be a full slice? Or do you want to still wrap the returns in Base.Slice?

I think it might be good to still wrap the return in `Base.Slice` because then things like `getindex(A, indices(...))` can know that it represents and inbounds collection.

chriselrod:
I think this looks good. Are you happy with it?

Tokazama:
With this final change I think it's good. Now you can use `indices` like `eachindex` but it will give you a slice so you now it spans the fully array of each.

src/ranges.jl
```diff
@@ -0,0 +1,231 @@
...
+  end
+end
+
+function indices(x::Tuple)
+  inds = map(eachindex, x)
+  @assert all(isequal(first(inds)), Base.tail(inds)) "Not all specified axes are equal: $inds"
+  return reduce(_pick_range, inds)
+end
+
+indices(x, d) = indices(axes(x, d))
```
timholy, line 209: This computes the indices of the indices, rather than the indices of `x`. You're implicitly assuming that `axes(x)` is idempotent. The older implementations of OffsetArrays would break this, for example, although the community seems to have widely settled on idempotency as a useful characteristic for the axes.

Tokazama: In terms of iterating values it should theoretically be the same, but it's not entirely idempotent because `indices(x::OneTo) = Slice(OptionallyStaticUnitRange(Val(1), last(x)))`. If `indices(x::OffsetArray, d) = indices(axes(x, d))` doesn't produce something that is appropriately offset then it could do something like `indices(x::OffsetArray, d) = indices(range(x.offset[d], stop = size(x, d))`.

PR #65, opened 2020-08-28
Fix stack overflow bug in known_length

Tokazama:
Added default `known_length(::Type{T}) where {T} = nothing`

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterface"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "2.12.0"
+version = "2.12.1"
 
 [deps]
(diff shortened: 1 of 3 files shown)


Tokazama:
Note that I didn't dig into wrapped parent arrays because that gets a bit tricky if we have something like a subarray where a particular dimension has it's length altered. We could alternatively have something that calculates the known_length based on the product of lengths from each dimensions's indices. Something like...

```julia
function axes_type(x, i)
    ...
end

known_length(::Type{T}) where {T} = prod(ntuple(i -> axes_type(T, i), Val(ndims(T))))
```

Tokazama:
I went ahead and just added support for `SubArray`. This meant adding `known_length(::Type{<:Number}) = 1` which is consistent with `length(::Number)`.

Note I also gave a patch bump to the version

chriselrod:
Good to merge and issue a new release?

PR #67, opened 2020-08-31
Fix #53 with ismutable default

### LINKED ISSUE #53: Create checks to avoid a StackOverflowError
Hi!

I was using a structure inside a workspace for OrdinaryDiffEq and I was getting this error:

```
ERROR: StackOverflowError:
Stacktrace:
 [1] ismutable(::Type{T} where T) at /Users/ronisbr/.julia/packages/ArrayInterface/YFV07/src/ArrayInterface.jl:19 (repeats 79978 times)
```

It seems that something is not defined inside ArrayInterface. I can reproduce it with the following MWE:

```julia
julia> mutable struct A
       a
       end

julia> t = A(1)

julia> ArrayInterface.ismutable(t)
ERROR: StackOverflowError:
Stacktrace:
 [1] ismutable(::Type{T} where T) at /Users/ronisbr/.julia/packages/ArrayInterface/YFV07/src/ArrayInterface.jl:19 (repeats 79978 times)
```

Question: can we define a fallback function that prints an error when we do not know if the type is or isn't mutable?

Tokazama:
This makes it so that `ismutable` has a default now, which is to use the `mutable` field of a type.

### DIFF
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -72,9 +72,10 @@ https://github.com/JuliaDiffEq/RecursiveArrayTools.jl/issues/19.
 ismutable(x) = ismutable(typeof(x))
 
 ismutable(::Type{<:AbstractArray}) = true
-ismutable(::Type{<:Number}) = false
 ismutable(::Type{<:AbstractRange}) = false
-ismutable(::Type{<:Tuple}) = false
+ismutable(::Type{<:AbstractDict}) = true
+ismutable(::Type{<:Base.ImmutableDict}) = false
+ismutable(::Type{T}) where {T} = T.mutable
 
 # Piracy
 function Base.setindex(x::AbstractArray,v,i...)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
I'm not sure we should do this. It's almost guaranteed to not be correct for wrapper types.

Tokazama:
It doesn't change any of the current behavior for wrapper types. It simply provides a fall back. If we want to account for wrappers I could replace the current `ismutable(::Type{<:AbstractArray}) = true` with

```julia

function ismutable(::Type{T}) where {T}
    if parent_type(T) <: T
        return T.mutable
    else
        return ismutable(parent_type(T))
    end
end
```

...and then we could explicitly define the mutability of each type.

Tokazama:
I added the recursive version that searches for a parent structure. I still haven't changed any of the tests, because the behavior is essentially the same.

Tokazama:
@ChrisRackauckas, is this PR still problematic?

chriselrod (approved):

ChrisRackauckas:
I think that works.

ChrisRackauckas:
Could it get a test? Fake wrapper

Tokazama:
So instead of testing a fake wrapped array I just tested views of StaticArrays. Turns out this actually fixes views on immutable arrays.

ChrisRackauckas (approved):

ChrisRackauckas:
I'm convinced.

Issue #66, opened 2020-08-30
Should we have pop, push and friends?


Tokazama:
StaticArrays has all of these non-mutating versions of typically mutating methods implemented, which might be a good fit for this package since we have `setindex`. It would be nice if these were available outside of a dependency on StaticArrays and part of the general array interface.

chriselrod:
Out of curiosity, what do you think of [BangBang.jl](https://github.com/JuliaFolds/BangBang.jl)?

It maybe nice to try and support a more uniform interface, e.g. `x = push!!(x, 2)`.
Of course, the methods changing the data structure's length won't be type stable for anything typed by size.

I hope @tkf doesn't mind getting pinged here, in case he'd like to give feedback or suggestions on a general interface.
Also because I wonder if this would be somewhat of a fracturing of these sorts of methods between ArrayInterface and BangBang.

Tokazama:
Ive always liked the syntax of BangBang but have never ended up using it. It might become more convenient as Julia gets more immutable collections though.

I also don't see what other possible meaning that syntax could reasoanably have, so it seems like a pretty safe thing to commit to if we moved forward with it.

PR #68, opened 2020-09-08
Non-mutating versions of pop, popfirst, etc. (#66)

Tokazama:
Some of this would be a lot cleaner if the `Static` type from #61 were available, but it's at least a start at creating generic versions for #66.

### DIFF
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -551,14 +551,12 @@ Return a new instance of `collection` with `item` inserted into at the given `in
 Base.@propagate_inbounds function insert(collection, index, item)
   @boundscheck checkbounds(collection, index)
   ret = similar(collection, length(collection) + 1)
-  @inbounds for i in indices(ret)
-    if i < index
+  @inbounds for i in firstindex(ret):(index - 1)
       ret[i] = collection[i]
-    elseif i == index
-      ret[i] = item
-    else
+  end
+  @inbounds ret[index] = item
+  @inbounds for i in (index + 1):lastindex(ret)
       ret[i] = collection[i - 1]
-    end
   end
   return ret
 end
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -543,6 +543,79 @@ function restructure(x::Array,y)
   reshape(convert(Array,y),size(x)...)
 end
 
+"""
+    push(collection, item)
+
+Return a new instance of `collection` with `item` inserted on the end.
+"""
+push(collection, x) = vcat(collection, x)
+
+"""
+    pushfirst(collection, item)
+
+Return a new instance of `collection` with `item` inserted at the beginning.
+"""
+pushfirst(collection, x) = vcat(x, collection)
+
+"""
+    insert(collection, index, item)
+
+Return a new instance of `collection` with `item` inserted into at the given `index`.
+"""
+Base.@propagate_inbounds function insert(collection, index, item)
+  @boundscheck checkbounds(collection, index)
+  ret = similar(collection, length(collection) + 1)
+  @inbounds for i in indices(ret)
+    if i < index
+      ret[i] = collection[i]
+    elseif i == index
+      ret[i] = item
+    else
+      ret[i] = collection[i - 1]
+    end
+  end
+  return ret
+end
+
+"""
+    pop(collection)
+
+Return a new instance of `collection` with the last item removed.
+"""
+@inline function pop(collection)
+  inds = pop(indices(collection))
+  return @inbounds(getindex(collection, inds))
+end
+
+"""
+    popfirst(collection)
+
+
+Return a new instance of `collection` with the first item removed.
+"""
+@inline function popfirst(collection)
+  inds = popfirst(indices(collection))
+  return @inbounds(getindex(collection, inds))
+end
+
+"""
+    deleteat(collection, index)
+
+Return a new instance of `collection` with the item at the given `index` removed.
+"""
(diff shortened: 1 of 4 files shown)


chriselrod:
I can split off the `Static` changes into a separate PR if you want.

Tokazama:
That would be great. Thanks

Tokazama:
Two concerns at this point:
1. I have only included generic definitions of these and haven't copied over anything defined in StaticArrays. Should I copy those methods over here but for the ArrayInterface version of the methods? Alternatively, this may be a good point to start getting other packages to depend on ArrayInterface directly instead of using Requires.
2. Are the generic versions acceptable?

chriselrod:
> Alternatively, this may be a good point to start getting other packages to depend on ArrayInterface directly instead of using Requires.

I'd be in favor of that.

> Are the generic versions acceptable?

Out of curiosity, what is the use case for these non-mutating functions?

`pop!` and `popfirst!` return the element you "pop off" the collection, so that one could write a program by pushing tasks/items to a vector, and then popping off work items.
That is, I thought returning the removed value was an important/convenient aspect of the API missing here.

Also, for `AbstractVector`s, an optimization might be to have `pop` return views.

If static sizing is involved, then type instability issues will make efficient use of such methods difficult.

Tokazama:
My initial goal here was to present the most generic version of those methods previously defined in StaticArrays as talking points. Here's a more opinionated breakdown:

* `insert`/`deleteat` have a similar utility as `setindex` (i.e. sometimes you want to preserve the original collection but have a separate instance given the mutating outcome). I think these can be pretty useful to have around. I'd be in favor of creating specific methods for tuples if others feel good about including these functions.
* `pushfirst`/`push`: I think this is intended to be a type stable version of `vcat(item, colletion)`/`vcat(collection, item)` for static vectors. However, I don't see why someone can't just define a type stable version of `vcat` instead given `item` should be  a known length of 1. It's not even shorter syntax so I'm not sure what the merit really is.
* `popfirst`/`pop`: The benefit of this one is that the equivalent type stable version available right now would probably be `x[(firstindex(x) -  Static(1)):lastindex(x)]`/`x[firstindex(x):(lastindex(x) + Static(1))]`. Which is a bit tedious to type. I've used this a couple of times, but TBH I typically only use `pop!`/`popfirst!` methods in interactive programming where type stability isn't important so `x[begin:end-1]` is probably just as good. I think you touched on the biggest issue that the non mutating version can't produce the "popped" element and the clipped collection in a similar way as the mutating version does. I'd actually like to see `Base.front` and `Base.tail` expanded to support ordered collections instead of this approach.

Tokazama:
I deleted the push/pop related methods for now, as I'm not sure if it's particular helpful to have them at this point. I also added support for tuples on both the `deleteat` and `insert`.

I happened upon some bugs with `OptionallyStaticUnitRange` and while fixing them cleaned up the code a bit. I don't anticipate any problems resulting from those changes but I can revert some of it if it could cause some friction.

Tokazama:
Should I move the changes to `OptionallyStaticUnitRange` and `Static` to a separate PR so that those fixes/changes can move forward?

src/ArrayInterface.jl
```diff
@@ -543,6 +543,105 @@ function restructure(x::Array,y)
...
 
+"""
+    insert(collection, index, item)
+
+Return a new instance of `collection` with `item` inserted into at the given `index`.
+"""
+Base.@propagate_inbounds function insert(collection, index, item)
+  @boundscheck checkbounds(collection, index)
+  ret = similar(collection, length(collection) + 1)
+  @inbounds for i in indices(ret)
```
chriselrod, line 554: I would split this into separate loops. The compiler may do so automatically, but I'd rather not rely on it.
Something like:
```julia
@inbounds for i in firstindex(ret):index-1
    ret[i] = collection[i]
end
ret[index] = item
@inbounds for i in index+1:lastindex(ret)
    ret[i] = collection[i-1]
end
```
to make sure it SIMDs well if possible.

src/static.jl
```diff
@@ -33,10 +33,13 @@ end
 Base.promote_rule(::Type{<:Static}, ::Type{<:Static}) = Int
 Base.:(%)(::Static{N}, ::Type{Integer}) where {N} = N
 
+Base.eltype(::Type{T}) where {T<:Static} = Int
 Base.iszero(::Static{0}) = true
 Base.iszero(::Static) = false
 Base.isone(::Static{1}) = true
 Base.isone(::Static) = false
+Base.zero(::Type{T}) where {T<:Static} = Static{0}()
```
chriselrod, line 41: Maybe we even want to define aliases like `const Zero = Static{0}()`?

chriselrod (approved):
I think this looks good. If you're more or less ready, I can go ahead and merge to save you from splitting off the `Static` changes.

Tokazama:
I looked at the `@code_native` output and I think it's working, but TBH I'm still getting used to reading it. I also added the constants `Zero` and `One`.

src/static.jl
```diff
@@ -6,6 +6,10 @@ Use `Static(N)` instead of `Val(N)` when you want it to behave like a number.
 struct Static{N} <: Integer
     Static{N}() where {N} = new{N::Int}()
 end
+
+const Zero = Static{0}()
+const One = Static{1}()
```
chriselrod, line 11: I said `const Zero = Static{0}()`, but what I meant was `const Zero = Static{0}`, so that we can use `::Zero` or `::One` in method signatures.

Tokazama:
Thanks for the feedback! Hopefully, anything else?

PR #69, opened 2020-09-09
Added tests for Static.

chriselrod:
@Tokazama 
This is the code for `Static`. The `OptionallyStaticUnitRange` could probably be cleaned up further.

Currently `known_*` functions still return `nothing` or values, but I added `static_*` that use these and return either `Static` numbers (if `known_` returns a value) or values (if they return `nothing`). These will likely often be more convenient to use, and seem strictly more powerful.

### DIFF
--- a/src/ranges.jl
+++ b/src/ranges.jl
@@ -90,8 +90,8 @@ struct OptionallyStaticUnitRange{T <: Integer, F <: Integer, L <: Integer} <: Ab
   end
 end
 
-Base.:(:)(L, ::Static{U}) where {U} = OptionallyStaticUnitRange(L, Static(U))
-Base.:(:)(::Static{L}, U) where {L} = OptionallyStaticUnitRange(Static(L), U)
+Base.:(:)(L::Integer, ::Static{U}) where {U} = OptionallyStaticUnitRange(L, Static(U))
+Base.:(:)(::Static{L}, U::Integer) where {L} = OptionallyStaticUnitRange(Static(L), U)
 Base.:(:)(::Static{L}, ::Static{U}) where {L,U} = OptionallyStaticUnitRange(Static(L), Static(U))
 
 Base.first(r::OptionallyStaticUnitRange) = r.start
@@ -142,6 +142,7 @@ end
 end
 
 @inline _try_static(::Static{N}, ::Static{N}) where {N} = Static{N}()
+@inline _try_static(::Static{M}, ::Static{N}) where {M, N} = @assert false "Unequal Indices: Static{$M}() != Static{$N}()"
 function _try_static(::Static{N}, x) where {N}
     @assert N == x "Unequal Indices: Static{$N}() != x == $x"
     Static{N}()
--- a/src/ranges.jl
+++ b/src/ranges.jl
@@ -44,6 +44,8 @@ known_step(::Type{<:AbstractUnitRange{T}}) where {T} = one(T)
 
 _get(x) = x
 _get(::Val{V}) where {V} = V
+_get(::Static{V}) where {V} = V
+_get(::Type{Static{V}}) where {V} = V
 _convert(::Type{T}, x) where {T} = convert(T, x)
 _convert(::Type{T}, ::Val{V}) where {T,V} = Val(convert(T, V))
 
@@ -56,7 +58,7 @@ at compile time. An `OptionallyStaticUnitRange` is intended to be constructed in
 from other valid indices. Therefore, users should not expect the same checks are used
 to ensure construction of a valid `OptionallyStaticUnitRange` as a `UnitRange`.
 """
-struct OptionallyStaticUnitRange{T,F,L} <: AbstractUnitRange{T}
+struct OptionallyStaticUnitRange{T <: Integer, F <: Integer, L <: Integer} <: AbstractUnitRange{T}
   start::F
   stop::L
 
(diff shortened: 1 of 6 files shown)


Tokazama:
This is great! The `static_*` methods seem like a nice way to ensure that static numbers are only used when asking for them.

Do you think it would be useful to support other integer types? If so I'm thinking of something that's still limited to types defined in base like one of the following

```julia
struct Static{N} <: Integer
    function Static{N}()  where {N}
        @assert isprimitivetype(typeof(N)) "$N is not a primitive type"
        return new{N}()
    end
end
```

```julia
struct Static{N} <: Integer
    function Static{N}()  where {N}
        @assert isa(N, Base.BitIntegerType) "$N is not a primitive type"
        return new{N}()
    end
end
```

chriselrod:
The reason I restricted it to `Int` was because of defining all the methods like `Base.:(::Static{0}, x) = Static{0}()`.
I also realize now I should also add more generic methods (i.e., where `x` is untyped like in that example). 

I also have a lot of trust that dispatch will be type stable.
But branches should be fine, too. If we go for allowing more integer types, we can use branch definitions such as:
```julia
@inline function Base.:(x::Static{M}, y) where {M}
    if iszero(M)
        x
    elseif isone(M)
        y
    else
        M * y
    end
end
```
and I think it's unlikely inference will fail.

timholy:
Allowing other Integer subtypes risks increasing your compile-time latency and likelihood of difficult-to-track-down bugs; try debugging why
```
julia> Val(3)
Val{3}()

julia> Val(Int16(3))
Val{3}()

julia> typeof(Val(3))
Val{3}

julia> typeof(Val(Int16(3)))
Val{3}
```
aren't the same type. Unless there's a compelling reason, diversifying the Integer types seems like a lot of pain for little or no benefit.

chriselrod:
I'll revert adding support for more integer types.

Tokazama:
I'm totally on board with just supporting `Int`. I just wanted to ensure it wouldn't be too restrictive for some future use.

chriselrod:
Just realized I also accidentally created a second branch called "static" where I added support for the integer types:
https://github.com/chriselrod/ArrayInterface.jl/tree/static
That one at chriselrod/ArrayInterface instead of at SciML/ArrayInterface like this PR.

Tokazama:
> An advantage of the branchy style worth considering is that it is less prone to ambiguity errors.

It might be good for us to start using `detect_ambiguities` to weed these things out. If it doesn't affect performance and reduces ambiguities the branching style may be preferable, but there is something beautiful in the simplicity of all those one line methods.

chriselrod:
I have noticed that using a lot of `@eval` loops can severely impact package load time, so I think it is normally preferable to keep the number of methods relatively small.

We have 36 definitions in the `+`/`-`/`*` loop.
The main savings we get is that, by not defining specific `::Static{0}` or `::Static{1}` definitions, we don't need to define any methods for the various combinations to avoid ambiguities.

I added
```julia
@test isempty(detect_ambiguities(ArrayInterface))
@test isempty(detect_unbound_args(ArrayInterface))
```

timholy:
Can also use https://github.com/JuliaTesting/Aqua.jl

Tokazama:
> Can also use https://github.com/JuliaTesting/Aqua.jl

That's really cool! I should be using that.

chriselrod:
Sure, now using `Aqua.jl`.

timholy:
@johnnychen94 introduced me to Aqua, I haven't yet used it.

@chriselrod, interesting. What does the `T >: Nothing` accomplish? Could you declare these
```julia
promote_rule(::Type{Nothing}, ::Type{S}) where S<:Static = ...
```

chriselrod:
> What does the T >: Nothing accomplish?

Resolves a type ambiguity on master.

> Could you declare these

I already had definitions like
```julia
promote_rule(::Type{Nothing}, ::Type{<:Static}) = ...
```
Is there a difference in specificity?
I still have the same ambiguity errors.

timholy:
Try rebuilding master? An ambiguity-related commit just got reverted within the last hour or so.

chriselrod:
https://github.com/JuliaLang/julia/pull/37484

To what extant is it better to err on the side of conservative here?

> Anyway, the two former ones seem serious enough that is worth to revert this to get back to the "baseline" and then eventually put it back when the problems have been fixed.

Sounds like it would be better long term to be strict.

chriselrod:
That said, no errors on Travis on 1.2, 1.5, or master now.
Presumably Travis hasn't updated master to reflect the reversion of the ambiguity commit yet (but I've tested locally before and after to confirm).

src/ranges.jl
```diff
@@ -193,7 +203,7 @@ specified then indices for visiting each index of `x` is returned.
 """
 @inline function indices(x)
   inds = eachindex(x)
-  if inds isa AbstractUnitRange{<:Integer}
+  if inds isa AbstractUnitRange#{<:Integer} # prevents inference
```
timholy, line 206: Could maybe use `inds isa AbstractUnitRange && eltype(inds) <: Integer`?

timholy (approved):
I didn't check in gory detail, but LGTM

src/ranges.jl
```diff
@@ -44,6 +44,8 @@ known_step(::Type{<:AbstractUnitRange{T}}) where {T} = one(T)
 
 _get(x) = x
 _get(::Val{V}) where {V} = V
```
Tokazama, line 46: Do we still need to support `Val` now that we're switching over to `Static` and `Val` can't be any of the fields in `OptionallyStaticUnitRange` anymore?

chriselrod: I've removed it.

src/ranges.jl
chriselrod: Done, and inference tests pass.

Tokazama:
I think this looks great.

PR #70, opened 2020-09-14
Change the indexing pipeline

Tokazama:
This PR presents a new pipeline for processing user provided indexing arguments before passing the resulting indices off to efficient native indexing code.
This approach was originally inspired throughout development of AxisIndices but evolved to address several other issues I discovered.
This exact iteration of the pipeline hasn't been fully tested yet because this one replaces custom more specialized methods I used in AxisIndices with recently developed and generic ones in ArrayInterface.
However, I've been using this general approach comfortably for months on basic array stuff and even some custom table and graph types.

# Changes and their motivation 

1. The biggest change is that `to_index(array, arg)` is now `to_index(axis, arg)`. In other words, the array being indexed isn't passed at this step.
  Instead, the indexing arguments are passed to `to_index` with the axes of the dimensions they correspond to.
  Although, `to_index` isn't exported from base it is still documented and formally a part of the array interface so I believe this would a breaking change and need to be in Julia 2.0 (I could be mistaken though).
  Also, bounds checking happens at the index level so that `checkindex` happens in `to_index`.
2. If there are any multidimensional arguments passed to `getindex` (`CartesianIndices`, `BitArray`, etc,) an initial pass to `flatten_args` is tries to break them down along each dimension.
  Once this step is done the only things in Base that still corresponds to multiple axes are vectors of `CartesianIndex` and multidim arrays of `Bool`.
  This greatly simplifies the code for `to_indices` and `to_axes`.
3. `to_axes` and `to_axis` are the other side of `to_indices` and `to_index`.
  They reconstruct axes for the new collection instead of using [`index_shape`](https://github.com/JuliaLang/julia/blob/master/base/multidimensional.jl#L757).
4. Once indexing arguments are converted to proper indices for an array the final point where we can alter the pipeline is in determining retrieval of the arrays individual elements and potential reconstruction of an array.
  This is done using the `UnsafeGetIndex` trait which the index at each dimension a trait and combines them (similar to what `BroadcastStyle` does with each collection that is broadcast over).
  Here, this only results in directions to retrieve a single value or construct a new array.

I should note that each of these changes helps with flexibility and many cases reduces the risk of ambiguities across packages.
For example, if one wants to specialize on each axis in base they need to either have a special array that doesn't use `to_indices` or they need to copy most of the `to_indices` methods in base and do something like...

```julia
to_indices(A::AbstractArray, axes::Tuple{<:MyCustomAxis,Vararg{Any}}, args::Tuple{Arg,Vararg{Any}})
```

...where there are several methods that are unique based on `Arg` alone.
Instead there are multiple spots that could function as dedicated hooks for unique behavior.
The goal with each "hook" was to identify a method at each step in the indexing pipeline and dispatch on as few positional arguments as possible.

# Potential future developments

## Generic support for unique axes

Using offset indices as an example, we assume in the following `OffsetAxis` and `CenteredAxis` have been previously defined.

```julia
abstract type OffsetStyle <: IndexStyle end

# an axis with a designated offset from 1
struct IndexOffset <: OffsetStyle end
IndexStyle(::Type{T}) where {T<:OffsetAxis} = IndexOffset()

# an axis that is always centered around 0
struct IndexCentered <: OffsetStyle end
IndexStyle(::Type{T}) where {T<:CenteredAxis} = IndexCentered()

function to_index(::OffsetStyle, axis, arg::Integer)
    index = Int(arg - sdoffset(axis))
    @boundscheck !if checkindex(Bool, parentindices(axis), index)
        throw(BoundsError(axis, arg))
    end
    return index
end
function to_index(::OffsetStyle, axis, arg::AbstractVector)
    index = arg .- sdoffset(axis)
    @boundscheck !if checkindex(Bool, parentindices(axis), index)
        throw(BoundsError(axis, arg))
    end
    return index
end

function to_axis(::IndexOffset, axis, arg, inds)
  return OffsetAxis(sdoffset(axis), to_axis(parentindices(axis), arg, inds))
end

function to_axis(::IndexCentered, axis, arg, inds)
  return OffsetAxis(to_axis(parentindices(axis), arg, inds))
end

```

This example illustrates an important benefit of being able to easily have unique behavior along each axis.
If we have a collection of elements and can access it through something like...
```julia
struct DArray{T,N,D,Axs} <: AbstractArray{T,N}
    data::D
    axes::Axs
end
```
...then once we implement a new axis type we don't need to reimplement all the related array methods to ensure something like `LinearAlgebra` supports offset indices.
This doesn't just apply to offset indices.
Padding like that found in ImageFiltering can be supported with an axis that redirects arguments to map to padded region ([reflect, circular, symmetric padding](https://juliaimages.org/latest/function_reference/#ImageFiltering.imfilter))
All static array types could be supported by simply replacing `StaticArray.SOneTo` with `SOneTo{L} = OptionallyStaticUnitRange{Static{1},Static{L}}` and using it as an axis.
In addition to automatically supporting  mixed axis types (like HybridArrays.jl), this also starts to solve the problem of composability for StaticArrays.
Instead of making a bunch of aliases for `DArray`, something like `DArray(::Tuple, ::Static, ::Static)` would produce something that is functionally equivalent to `SMatrix` and `DArray(::Vector, ::Static, ::Static)` would be like a `SizedMatrix`.

It is obviously not the goal of ArrayInterface to define everything and one array to rule them all.
Instead, this is a step towards making it so that if someone creates a PaddedAxes.jl package then padded axes could theoretically work for distributed/shared/dense/etc arrays that adopt this interface.

## Points to safely hook into the array interface

A lot of it is motivated by what was discussed at the BOF for fancy array indexing at JuliaCon this year where Chris Rackauckas pointed out that what we really want is a consistent API for mapping arguments from the square brackets to whatever is physically in memory.
That got me thinking about how many decisions points we have to step into the code and change the outcome before reaching the final `unsafe_load`.
Furthermore, can we do it by focusing on a single argument position so that we can avoid ambiguities.
Here are some of the ones this pipeline makes readily available:

* `getindex(A, args...)`: dispatch on `A` to change everything
* `to_indices(A, args::Tuple)`:  dispatch on `A` to change how indices are produced
* `to_index(::IndexStyle, axis, arg)`: dispatch on `IndexStyle` of `axis` to change how an index is produced
* `to_axes(A, args::Tuple, inds::Tuple)`: dispatch on `A` to change how the new axes/shape of a collection is made.
* `to_axis(::IndexStyle, axis, arg, index)`: dispatch on `IndexStyle` of `axis` to change how the new axis is constructed
* `UnsafeGetIndex(indices)`: dispatch on the final indices produced from `to_indices` to redirect what method is used to get values from `A`
* `flatten_args(A, axs::Tuple, args::Tupl{Arg,Vararg{Any}})`: `Arg` could can be used for multidimensional expressions (like that in EllipsNotation.jl).

## `similar`

One of the problems with methods like similar is the tuple of arguments for the size/indices causes unexpected ambiguities and sometimes returns a different array type. Adoption of this axis centric arrays could help handle that by instead having something like.

```julia
function similar(A, ::Type{T}, inds...)
    axs = to_new_axes(A, axes(A), (), inds)
    return unsafe_reconstruct(A, Vector{T}(undef, prod(map(length, axs)), axs), axs)
end
```

# Remaining problems

I haven't really come up with an approach I like to make the final step generic.
For returning a single element we can always to something where we require specifying some method like `load_array_element(A, ii::Int...)`.
Returning new collections is a bit more tricky because even if we can't necessarily use `similar` if we have an immutable array.
Also, we probably want something generic that can point to optimizations like those in `LoopVectorization.jl` (e.g., `fuse_indices_and_store!(new_array, old_array, inds)`).


src/indexing.jl
```diff
@@ -0,0 +1,231 @@
+
+_maybe_tail(::Tuple{}) = ()
+_maybe_tail(x::Tuple) = tail(x)
+
+#
+#    to_indices(A, args)
+#
+function to_indices end
+const to_indices = ArrayInterface.to_indices
```
chriselrod, line 9: Redundant?

Tokazama: So that all `to_indices` methods in the module go to the one specified here instead of the one in base.

chriselrod: Isn't the `function to_indices end` enough?
```julia
julia> function inv end
inv (generic function with 0 methods)

julia> inv(x) = 2x
inv (generic function with 1 method)

julia> Base.inv(3)
0.3333333333333333

julia> inv(3)
6
```

Tokazama: Oh, you're totally right. I was basing this off of what happens when exporting it into a new module where `const to_indices = ArrayInterface.to_indices` was necessary.

chriselrod:
Would be good to add tests here, perhaps a simple custom array type implementing this for indexing?

Tokazama:
This PR is still a little rough. I'll be updating it over the next day or two with some improvements (and of course tests).

I'm still not settled on the final "unsafe" part of the pipeline. I'm trying to take [this](https://github.com/JuliaLang/julia/blob/master/base/multidimensional.jl#L755) part of the indexing pipeline and formalize the access to raw data and reconstructing the array. `unsafe_get_element` should have have something like:

```julia
function unsafe_get_element(A::AbstractArray, args::Tuple, inds::Tuple)
    return unsafe_get_element(device(A), A, args, inds)
end
function unsafe_get_element(::CPUPointer, A::AbstractArray, args::Tuple, inds::Tuple)
    return unsafe_load(pointer(A), to_linear(A, inds))
end
```

But for something where we want to reconstruct the array it's not necessarily going to be a parent array (as is currently assumed in this PR). It seems like there needs to be another interface for accessing non array collections serve as the data underlying data. Perhaps the interface could be as simple as:

```julia
function unsafe_get_collection(A, args, inds)
    return unsafe_reconstruct(
        A,
        unsafe_get_data(data_parent(A), inds)
        to_axes(A, args, inds)
    )
end
```

Tokazama:
I think this is in a good place now. In the future I'd like to have `unsafe_get_collection` to advance a bit more to support immutable collections and fusing indices where possible, but I think this PR gets it to the point where indices are checked, correctly typed, and enough information is still there to make these things happen easily.

chriselrod:
If you're adding
```julia
Base.:(-)(r::OptionallyStaticRange) = -static_first(r):-static_step(r):-static_last(r)
```
it'd make sense to also add
```julia
Base.:(-)(r::OptionallyStaticUnitRange) = -static_first(r):StaticInt(-1):-static_last(r)
```

I'll also add these after this PR if you don't:
```julia
Base.reverse(r::OptionallyStaticRange) = static_last(r):-static_step(r):static_first(r)
Base.reverse(r::OptionallyStaticUnitRange) = static_last(r):StaticInt(-1):static_first(r)
```

Tokazama:
This actually takes care of `Base.:(-)(r::OptionallyStaticUnitRange) = -static_first(r):StaticInt(-1):-static_last(r)` because `OptionallyStaticRange` is a union of `OptionallyStaticUnitRange` and `OptionallyStaticStepRange`.

I added tests for `reverse` and it just works with what we have now.

src/indexing.jl
```diff
@@ -0,0 +1,502 @@
...
+
+Returns a collection of `A` given `inds`. `inds` is assumed to be bounds checked prior.
+"""
+function unsafe_get_collection(A, inds)
+    axs = to_axes(A, inds)
+    dest = similar(A, axs)
+    map(Base.unsafe_length, axes(dest)) == map(Base.unsafe_length, axs) || throw_checksize_error(dest, axs)
+    Base._unsafe_getindex!(dest, A, inds...) # usually a generated function, don't allow it to impact inference result
+    return dest
+end
```
Tokazama, line 414: Should this be replaced by:

```julia
unsafe_get_collection(A, inds) = unsafe_get_collection(device(A), A, inds)
function unsafe_get_collection(::CPUIndex, A, inds)
    axs = to_axes(A, inds)
    dest = similar(A, axs)
    map(Base.unsafe_length, axes(dest)) == map(Base.unsafe_length, axs) || Base.throw_checksize_error(dest, axs)
    Base._unsafe_getindex!(dest, A, inds...) # usually a generated function, don't allow it to impact inference result
    return dest
end

function unsafe_get_collection(::CPUPointer, A, inds)
    data = vload(pointer(A), unroll(A, inds))
    return unsafe_reconstruct(A, data; axes=to_axes(A, inds))
end
```

This would require formalizing some of the interface for raw access to memory, but it would pretty much make this complete.

Tokazama:
Any other opinions on the current implementation?

Tokazama:
Has this PR gotten too big to review? Should I move the other fixes to another PR?

src/ranges.jl
```diff
@@ -214,10 +222,17 @@ unsafe_length_one_to(::StaticInt{L}) where {L} = L
...
     else
-        return Base.checked_add(Int(div(Base.checked_sub(rtart, stop), -step)), 1)
+        return Base.checked_add(Int(div(Base.checked_sub(start, stop), -step)), 1)
     end
 end
 
+@propagate_inbounds function Base.getindex(r::OptionallyStaticUnitRange, s::AbstractUnitRange{<:Integer})
+    @boundscheck checkbounds(r, s)
+    f = static_first(r)
+    f = f - one(f)
```
chriselrod, line 232: I'd rename `f` in case of type instabilities. This shows up in `@code_warntype` for me.

test/indexing.jl
```diff
@@ -0,0 +1,130 @@
...
+    @test cartesian[linear] == cartesian
+    @test cartesian[vec(linear)] == vec(cartesian)
+    @test cartesian[cartesian] == cartesian
+    @test cartesian[vec(cartesian)] == vec(cartesian)
+    @test linear[2:3] === 2:3
+    @test linear[3:-1:1] === 3:-1:1
+    @test_throws BoundsError linear[4:13]
+end
+
+@testset "3-dimensional" begin
```
chriselrod, line 97: Mind adding some `@inferred` for good measure? Sometimes inference can hit heuristic limits.

chriselrod (commented):
Overall, I think it looks good.
When I rewrite PaddedMatrices, I'll try defining 
```julia
Base.getindex(A::AbstractStrideArray, i...) = ArrayInterface.getindex(A, i...)
ArrayInterface.unsafe_get_element(A::AbstractStrideArray, inds) = ...
```
That would also rely on `Base._getindex` being defined appropriately for collections to work.
I do want to override some of the normal behavior (I want slicing to produce views by default) anyway, but it would be convenient if we could minimize the methods someone has to define for a custom array type to opt into this.

I'll likely have more comments after experimenting with it in that manner, which will be several weeks away.
But I'm fine with merging it after addressing the small handful of comments below.

chriselrod (approved):
Looks good -- ready to merge?

Tokazama:
> I do want to override some of the normal behavior (I want slicing to produce views by default) anyway, but it would be convenient if we could minimize the methods someone has to define for a custom array type to opt into this.

This is the point behind `UnsafeIndex`, a last step for redirecting the entire pipeline. I've been using it to add padding along each dimension, but I didn't consider that the array type might be important too. Should these be changed to act more like `argdims` where an array is also passed to `UnsafeIndex`

chriselrod:
> I've been using it to add padding along each dimension, but I didn't consider that the array type might be important too. Should these be changed to act more like argdims where an array is also passed to UnsafeIndex

I went ahead and merged the PR, but I think that'd be useful before the next release.

PR #74, opened 2020-09-24
OptionallyStaticStepRange

Tokazama:
This PR implements the version of `OptionallyStaticUnitRange` when steps aren't 1. The code is considerably more concise than what would have been required by the original attempts to implement it because we have now focused on only using `Int` as the static type we are supporting.


chriselrod:
Not a part of the PR, but I noticed
```julia
@inline function get_index_unit_range(r, i)
  val = first(r) + (i - 1)
  @boundscheck if (i < 1) || (val > last(r) && val < first(r))
    throw(BoundsError(r, i))
  end
  return convert(eltype(r), val)
end
```
The check `(val > last(r) && val < first(r))` is of course incorrect:
```julia
julia> using ArrayInterface: StaticInt

julia> r = 1:StaticInt(100)
1:StaticInt{1}():StaticInt{100}()

julia> r[200]
200
```
You probably just want `val > last(r)`

chriselrod:
If you don't want to combine the two implementations (so that `OptionallyStaticUnitRange` can still subtype `AbstractUnitRange`), perhaps we should have `OptionallyStaticRange(start, ::StaticInt{1}, stop)` construct an `OptionallyStaticUnitRange` instead?

Tokazama:
> We probably just want val > last(r)

You're right. Since this is really rounding out the range interface here I'll just include it here.

> If you don't want to combine the two implementations (so that OptionallyStaticUnitRange can still subtype AbstractUnitRange*), perhaps we should have OptionallyStaticStepRange(start, ::StaticInt{1}, stop) construct an OptionallyStaticUnitRange instead?

I think it would be odd to specify one type as a call but return a different type, but I think that's a good idea for the range constructor (`(:)(::Int, ::StaticIint{1}, ::Int`). Does that sound reasonable?

> *Would be great if everyone/everything just used traits, so that separate definitions to opt in wouldn't be needed!

Completely agree. Unfortunately all of base has the assumption that indices are `AbstractUnitRange` (e.g., can only construct `CartesianIndices` and `LinearIndices` with `AbstractUnitRange{Int}` subtypes).

The schema I have in my head is that `StaticInt` is really driving the public interface for this. `OptionallyStaticUnitRange` and `OptionallyStaticStepRange` are more of internal implementation. Hopefully someday we can just have `OptinallyStaticRange`.

chriselrod:
> I think it would be odd to specify one type as a call but return a different type, but I think that's a good idea for the range constructor `(:)(::Int, ::StaticIint{1}, ::Int)`. Does that sound reasonable?

I don't know why someone would want that, but that sounds reasonable, as long as it's for each signature (which you've already added).

chriselrod (approved):

Tokazama:
The constructors for `StaticInt`? I can fix those now.

Tokazama:
> > I think it would be odd to specify one type as a call but return a different type, but I think that's a good idea for the range constructor `(:)(::Int, ::StaticIint{1}, ::Int)`. Does that sound reasonable?
> 
> I don't know why someone would want that, but that sounds reasonable, as long as it's for each signature (which you've already added).
> 

Do you want me to remove this? I don't really care either way.

chriselrod:
I meant, I don't know why someone would want an `OptionallyStaticStepRange{<:Any,StaticInt{1},<:Any}` over an `OptionallyStaticUnitRange{<:Any,<:Any}`.

PR #75, opened 2020-10-10
Rework UnsafeIndex

Tokazama:
This PR is aimed at making the `UnsafeIndex` trait aware of the array that accompanies indexing arguments. I also replaced `IndexStyle` with a the new `ArrayStyle` trait for `argdims`. This allows unique indexing behavior while still allowing the use of `IndexLinear` to specify memory layout.

### DIFF
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -65,14 +65,12 @@ Base.promote_rule(::Type{X}, ::Type{Y}) where {X<:UnsafeIndex,Y<:UnsafeGetElemen
 @generated function UnsafeIndex(s::ArrayStyle, ::Type{T}) where {N,T<:Tuple{Vararg{<:Any,N}}}
     if N === 0
         return UnsafeGetElement()
-    elseif N === 1
-        return UnsafeIndex(s, T.parameters[1])
     else
-        out = typeof(UnsafeIndex(s, T.parameters[1]))
-        for i in 2:N
-            out = promote_type(out, typeof(UnsafeIndex(s, T.parameters[i])))
+        e = Expr(:call, promote_type)
+        for p in T.parameters
+            push!(e.args, :(typeof(ArrayInterface.UnsafeIndex(s, $p))))
         end
-        return out()
+        return Expr(:block, Expr(:meta, :inline), Expr(:call, e))
     end
 end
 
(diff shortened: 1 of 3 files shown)


src/indexing.jl
```diff
@@ -1,27 +1,85 @@
...
+    elseif N === 1
+        return UnsafeIndex(s, T.parameters[1])
+    else
+        out = typeof(UnsafeIndex(s, T.parameters[1]))
+        for i in 2:N
+            out = promote_type(out, typeof(UnsafeIndex(s, T.parameters[i])))
+        end
+        return out()
+    end
+end
```
chriselrod, line 77: If users are likely to add their own `UnsafeIndex` methods, being `@generated` is likely to cause problems if people want to define `UnsafeIndex` for their own `Type{T}`. But maybe that isn't a big deal, since defining it for their own Style is probably more likely?

Tokazama: > But maybe that isn't a big deal, since defining it for their own Style is probably more likely?

That' what I was originally assuming but it would be nice if defining a new instance of `UnsafeIndex` didn't require redefining that method too.

chriselrod:
I think your new `UnsafeIndex` looks good. I'll merge this and issue a new release of ArrayInterface if you think this is good to go.

ChrisRackauckas:
Hmm, did this PR introduce a breaking change?

https://travis-ci.org/github/SciML/OrdinaryDiffEq.jl/jobs/734958561#L762-L777

chriselrod:
Probably accidental, @Tokazama ?

The breaking change is because this PR introduced `ArrayInterface.setindex!` (and `ArrayInterface.getindex`) that added it's own indexing implementation.
That was from [this PR](https://github.com/SciML/ArrayInterface.jl/pull/70).
That means that
```julia
allowed_setindex!(x,v,i...) = setindex!(x,v,i...)
```
now suddenly refers to `ArrayInterface.setindex!` instead of `Base.setindex!`.

I'll make the simplest fix and specify
```julia
allowed_setindex!(x,v,i...) = Base.setindex!(x,v,i...)
```
We should consider if we want to make it `ArrayInterface.setindex!` and bump the major version or not.
The advantage of `ArrayInterface.setindex!` is that it's easier to customize indexing behavior on a per-axis basis.

so that the patch release isn't breaking. I should've at least bumped the minor version instead.

Tokazama:
Sorry, i didn't catch that potential issue. I'll double check for any other spots this may be an issue.

PR #83, opened 2020-11-09
Formalize interface for keyword arguments in indexing

Tokazama:
Currently there is no standard meaning for keyword arguments when passed to `getindex`. Often keyword arguments are propagated optional arguments for some method along the call chain. However, the most popular use of keyword arguments for indexing (as far as I'm aware) is that implemented by NamedDims.jl where dimension names are the keywords. As discussed in [this issue](https://github.com/invenia/NamedDims.jl/issues/147), this is problematic if the wrapper around a named dimension array alters the dimensions of an array (e.g., `SubArray`) because the indices associated with the keyword arguments passed on to subsequent calls don't necessary have the same meaning once the wrapper is removed (e.g., `getindex(x; kwargs...) = getindex(parent(x); kwargs...)`) .

In this PR I added a minimal interface for incorporating named dimensions in the indexing pipeline. This provides generic support for what is arguably the most widely used application of keyword arguments for array indexing. This still propagates kwargs when the array doesn't have any dimension names. Since there currently isn't generic propagation of kwargs for indexing this shouldn't break anything.

Somewhat orthogonal to the issue of formalizing the keywords component of the indexing interface, this also standardizes named dimensions. It would be really nice if users didn't have to bother with figuring out if the model they used returns a `NamedDims.NamedDimsArray`,  `DimensionalData.AbstractDimArray`, or `AxisArrays.AxisArray` when using NamedDims.jl, DimensionalData.jl, or AxisArrays.jl when indexing.

Some of the code is from NamedDims.jl but a lot of it still needs to be benchmarked and tested. I figured I'd wait to do that until I knew if this was a desirable addition and if people agreed on the interface.

@oxinabox , you may be interested in this.

### DIFF
--- a/src/dimensions.jl
+++ b/src/dimensions.jl
@@ -1,14 +1,4 @@
 
-Base.@pure function _permute_dimnames(x::Tuple{Vararg{Symbol,D}}, perm::NTuple{N,Int}) where {D,N}
-    ntuple(Val(N)) do i
-        if D < i
-            return :_
-        else
-            return getfield(x, getfield(perm, i))
-        end
-    end
-end
-
 """
     has_dimnames(x) -> Bool
 
@@ -30,28 +20,29 @@ end
 Return the names of the dimensions for `x`.
 """
 @inline dimnames(x) = dimnames(typeof(x))
-@inline dimnames(x, i::Int) = dimnames(typeof(x), i)
-@inline function dimnames(::Type{T}, d::Int) where {T}
-    if has_dimnames(T)
-        return getfield(dimnames(T), d)
+@inline dimnames(x, i::Integer) = dimnames(typeof(x), i)
+@inline dimnames(::Type{T}, d::Integer) where {T} = getfield(dimnames(T), to_dims(T, d))
+@generated function dimnames(::Type{T}) where {T}
+    if parent_type(T) <: T
+        return ntuple(i -> Symbol(:dim_, i), Val(ndims(T)))
     else
-        return nothing
+        return dimnames(parent_type(T))
     end
 end
-@inline function dimnames(::Type{T}) where {T}
-    return ntuple(i -> dimnames(parent_type(T), i), Val(ndims(T)))
-end
 @inline function dimnames(::Type{T}) where {T<:Union{Transpose,Adjoint}}
-    return map(d -> dimnames(dimnames(parent_type(T))), (2, 1))
+    return map(i -> dimnames(parent_type(T), i), (2, 1))
 end
 @inline function dimnames(::Type{T}) where {I,T<:PermutedDimsArray{<:Any,<:Any,I}}
-    return map(d -> dimnames(dimnames(parent_type(T))), I)
+    return map(i -> dimnames(parent_type(T), i), I)
 end
-@generated function dimnames(::Type{T}) where {I,T<:SubArray{<:Any,<:Any,<:Any,I}}
+function dimnames(::Type{T}) where {P,I,T<:SubArray{<:Any,<:Any,P,I}}
+    return _sub_array_dimnames(Val(dimnames(P)), Val(argdims(P, I)))
+end
+@generated function _sub_array_dimnames(::Val{L}, ::Val{I}) where {L,I}
     e = Expr(:tuple)
-    for i in 1:ndims(T)
-        if argdims(A, I.parameters[i]) > 0
-            push!(e.args, dimenames(parent_type(P), i))
+    for i in 1:length(L)
+        if I[i] > 0
+            push!(e.args, QuoteNode(L[i]))
         end
     end
     return e
@@ -63,7 +54,6 @@ end
 This returns the dimension(s) of `x` corresponding to `d`.
 """
 to_dims(x, d::Integer) = Int(d)
-to_dims(x, d::StaticInt) = d
 to_dims(x, d::Colon) = d   # `:` is the default for most methods that take `dims`
 @inline to_dims(x, d::Tuple) = map(i -> to_dims(x, i), d)
 @inline function to_dims(x, d::Symbol)::Int
--- a/src/dimensions.jl
+++ b/src/dimensions.jl
@@ -22,16 +22,19 @@ Return the names of the dimensions for `x`.
 @inline dimnames(x) = dimnames(typeof(x))
 @inline dimnames(x, i::Integer) = dimnames(typeof(x), i)
 @inline dimnames(::Type{T}, d::Integer) where {T} = getfield(dimnames(T), to_dims(T, d))
-@generated function dimnames(::Type{T}) where {T}
+@inline function dimnames(::Type{T}) where {T}
     if parent_type(T) <: T
-        return ntuple(i -> Symbol(:dim_, i), Val(ndims(T)))
+        return ntuple(i -> :_, Val(ndims(T)))
     else
         return dimnames(parent_type(T))
     end
 end
 @inline function dimnames(::Type{T}) where {T<:Union{Transpose,Adjoint}}
-    return map(i -> dimnames(parent_type(T), i), (2, 1))
+    return _transpose_dimnames(dimnames(parent_type(T)))
 end
+_transpose_dimnames(x::Tuple{Symbol,Symbol}) = (last(x), first(x))
+_transpose_dimnames(x::Tuple{Symbol}) = (:_, first(x))
+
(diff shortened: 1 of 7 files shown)


chriselrod:
Are you waiting for a comment from oxinabox, or is this ready?

Tokazama:
I'd like to see what @oxinabox , @nickrobinson251, or @mcabbot think so that I don't step on any toes by using some of the ideas generated in NamedDims.jl.

Tokazama:
@timholy, you were part of the original discussion surrounding `dimnames`. Does this seem reasonable to you?

src/indexing.jl
```diff
@@ -385,21 +385,34 @@ end
     ArrayInterface.getindex(A, args...)
 
 Retrieve the value(s) stored at the given key or index within a collection. Creating
-another instance of `ArrayInterface.getindex` should only be done by overloading `A`.
+anothe instance of `ArrayInterface.getindex` should only be done by overloading `A`.
```
timholy, line 388:
```suggestion
another instance of `ArrayInterface.getindex` should only be done by overloading `A`.
```

src/dimensions.jl
```diff
@@ -0,0 +1,157 @@
...
+    dimnames(x, d) -> Symbol
+
+Return the names of the dimensions for `x`.
+"""
+@inline dimnames(x) = dimnames(typeof(x))
+@inline dimnames(x, i::Integer) = dimnames(typeof(x), i)
+@inline dimnames(::Type{T}, d::Integer) where {T} = getfield(dimnames(T), to_dims(T, d))
+@generated function dimnames(::Type{T}) where {T}
+    if parent_type(T) <: T
+        return ntuple(i -> Symbol(:dim_, i), Val(ndims(T)))
```
timholy, line 27: As you know, I've had mixed experience with having default names (e.g., https://github.com/JuliaImages/ImageCore.jl/issues/112). Recently, though, you've played around with this a lot more than I have. Is your sense that this really does make life better?

timholy (commented):
From a quick skim it seems pretty reasonable to me, but I can't pretend to have given it a careful read. Just a couple of comments.

src/dimensions.jl
Tokazama: Would it make sense to do what NamedDims.jl does and return `:_` for unnamed dimensions? Even if we don't have default names for arrays that return false on `has_dimnames` then we have to come up with something for the transpose/adjoint of a vector or when creating a view of an array with extra slices (e.g., `ones(2,2)[:,:,:,:]`).

Another slightly related problem is this could lead to multiple dimensions named `:_`. It makes more sense to me to have dimensions with strictly unique names like a `NamedTuple`, but perhaps that's not the right way of thinking about it.

timholy: I confess I haven't had time to play with NamedDims, although in the (hopefully not too) long run I definitely want to transition JuliaImages to something that is actively maintained. Using a default, non-unique name seems like a pretty good solution to me. My own experience suggests that giving things unique but meaningless names doesn't seem likely to be very helpful, and might be hard to distinguish unambiguously from intentional names. (What if users deliberately choose `dim_1`, `dim_2`, etc? That would spoil any attempt to detect whether the names are meaningful or just default.)

mcabbott: Perhaps it's worth saying that the default name `:_` is a wild-card, which matches anything. As you say, generated names `:dim_1 ` etc. would probably need to be tagged as being 2nd-class somehow, so that clashes aren't taken as seriously as clashes of meaningful names.

I agree that the NamedTuple analogy is appealing. But I was persuaded that allowing even meaningful names to repeat is natural, as they arise from things like `m = v .+ v'`, or correlation matrices. In such cases `sum(m, dims=:i)` currently picks the first matching dimension, and picking one certainly seems like the right behaviour. (I'm not sure this is tested & guaranteed.) `m[i=1]` fixes both, but people like to say it's undefined, and should perhaps be made an error. The alternative would be some mechanism for always adding primes in such cases, I'm not sure this has been much explored. 

(Have not thought much about the broader proposal here, and whether this should live here not in NameDims.)

Tokazama: Okay. I'm convinced that that the old version with `:dim_i` naming only complicates the problem I was trying to solve. I think the wild card (`:_`) also helps avoid defining any formal rules about repeated dimension names here, and pushes that off as an implementation detail for packages like `NamedDims.jl`.

Tokazama:
Thanks @timholy and @mcabbott for taking time to provide feedback.

I'm not sure it's helpful to add much more complexity to dimensions or kwargs for the indexing interface here. Unless there are anymore suggestions I think this is ready.

Tokazama:
@chriselrod did you want me to start incorporating traits from #61 into methods like `_unsafe_getindex!`? I figured that could turn into a rather lengthy discussion, so I was saving it for another PR. But I'm open to whatever you think is best.

phlavenk:
Hi @Tokazama, I'd like to ask, if the interface also allows to interface `AxisArrays.jl`. What is the way to allow a dispatch specialized on handling of 1D array with an axis composed of equidistantly sampled time series in unitful's "Time" units (e.g. nanoseconds or seconds) and a different dispatch if the axis is in frequency domain (Unitful's Hz) with non-equidistant increments (typically log10, or a 1, 3, 10, 30, ... or 1, 2, 4, 8, .... )? And does the subarray slicing support intervals in some way? 
I'm not good in designing interfaces nor an expert on dispatch, but an example how to achieve the mentioned tasks using the proposed universal array interface would help me a lot to get confidence and migrate my code to this API.

chriselrod:
> @chriselrod did you want me to start incorporating traits from #61 into methods like _unsafe_getindex!? I figured that could turn into a rather lengthy discussion, so I was saving it for another PR. But I'm open to whatever you think is best.

It'd probably be better for a separate PR, focusing this one on keyword args

Tokazama:
> Everyone's fine with the `@pure` uses here?

BTW. These were copied directly from `NamedDims.jl` and I think Lyndon White (oxinabox) implemented them. I trust his judgement on these sorts of things, but if it's still a concern I could work out a separate solution.

@phlavenk, this is very much suited to work with something like AxisArrays.jl. You're comment motivated me to clean up some of my code in AxisIndices.jl (which has a lot of comments and mysteriously named inner functions right now). So here are some minimal examples of what could get this working with some of the stuff you mentioned. (Disclaimer: consider this pseudo code b/c I haven't actually tested it for AxisArrays specifically)

```julia
ArrayInterface.dimnames(::Type{T}) where {T<:AxisArray} = AxisArrays.axisnames(T)
ArrayInterface.has_dimnames(::Type{T}) where {T<:AxisArray} = true
function ArrayInterface.to_index(::IndexStyle, axis::Axis, arg::Interval)
    return findall(in(arg), axis.val)
end
Base.getindex(x::AxisArray, args...) = ArrayInterface.getindex(x, args...)
Base.getindex(x::AxisArray; kwargs...) = ArrayInterface.getindex(x; kwargs...)
function ArrayInterface.unsafe_get_element(x::AxisArray, inds)
    return @inbounds(parent(x)[inds...])
end
```

You can take advantage of this currently with `AxisIndices.NamedAxisArray` though.

chriselrod:
What's the status on this?

Tokazama:
I made one more change to conform with changes to strides. There's a fallback for `contiguous_axis` if it's known there is a parent array and `contiguous_axis` isn't directly defined.

```julia
function contiguous_axis(::Type{T}) where {T}
    if parent_type(T) <: T
        return nothing
    else
        return contiguous_axis(parent_type(T))
    end
end
```

If that's fine then this should be ready.

chriselrod:
Yeah, I think that's fine.
I'm wondering if we should assume that arrays inherit properties from `parent_type` to reduce the boiler plate for adding support to custom array types that sever mostly as wrappers for another array, with a few added features.

E.g., to add support for `HybridArray`s, all that ought to be needed is adding static size information, and automate translation of this to static strides through the dense dims trait.

Alternatively, we could add an "inherits from parent" trait to make this optional. That'd be a bit safer, and only slightly less convenient.
That'd be something for a separate PR of course.

Tokazama:
> I'm wondering if we should assume that arrays inherit properties from parent_type to reduce the boiler plate for adding support to custom array types that sever mostly as wrappers for another array, with a few added features.

It would be nice if this were more formally addressed so there isn't any ambiguity about how traits function with arrays. I'd be happy to help out with that discussion/PR.

PR #86, opened 2020-11-10
Fix contiguous_axis for PermutedDimsArray with discontiguous parent

jipolanco:
This fixes `contiguous_axis(A)` when `A` is a `PermutedDimsArray` with a discontiguous parent array.

Example:

```julia
u = zeros(3, 4, 5)
v = @view u[2, :, :]  # discontiguous subarray
A = PermutedDimsArray(v, (2, 1))

contiguous_axis(v)  # returns Contiguous(-1) as expected
contiguous_axis(A)  # BoundsError: attempt to access (2, 1) at index [-1]
```

After the changes, `contiguous_axis(A)` returns `Contiguous(-1)`.


chriselrod:
Thanks!

PR #87, opened 2020-11-10
Fix contiguous_axis_indicator for the case of unknown contiguity

### LINKED ISSUE #84: contiguous_axis_indicator for arrays with unknown contiguity
Currently, `contiguous_axis` returns `nothing` for an array type with unknown contiguity properties. However, for the same array type, `contiguous_axis_indicator` is undefined, as there is no defined `contiguous_axis_indicator(::Nothing, ::Val)` method.

Is this intentional? If not, would it make sense here to return `nothing`, just like for `contiguous_axis`?

Minimal example:

```julia
import ArrayInterface: contiguous_axis, contiguous_axis_indicator

# Define dummy array type with unknown contiguity properties
struct DummyArray{T,N} <: AbstractArray{T,N}
    dims :: Dims{N}
    DummyArray{T}(::UndefInitializer, dims...) where {T} = new{T,length(dims)}(dims)
end
Base.size(x::DummyArray) = x.dims
Base.getindex(::DummyArray{T}, inds...) where {T} = zero(T)

A = DummyArray{Float64}(undef, 3, 5)
contiguous_axis(A)            # returns nothing
contiguous_axis_indicator(A)  # MethodError: no method matching contiguous_axis_indicator(::Nothing, ::Val{2})
```

jipolanco:
Closes #84

### DIFF
--- a/src/stridelayout.jl
+++ b/src/stridelayout.jl
@@ -62,6 +62,7 @@ Returns a tuple boolean `Val`s indicating whether that axis is contiguous.
 """
 contiguous_axis_indicator(::Type{A}) where {D, A <: AbstractArray{<:Any,D}} = contiguous_axis_indicator(contiguous_axis(A), Val(D))
 contiguous_axis_indicator(::A) where {A <: AbstractArray} = contiguous_axis_indicator(A)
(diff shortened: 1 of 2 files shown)


chriselrod:
Thanks again!

PR #88, opened 2020-11-10
Improve handling of general SubArrays

### LINKED ISSUE #85: `size` fails for view with discontiguous slices
For array views, `size` only seems to work when the slices are contiguous (of the form `:` or `a:b`). When there is a step (as in `2:2:10`), the associated dimension is lost, which I guess is not intended.

Example:

```julia
u = rand(4, 6)
v = @view u[1:3, :]
w = @view u[2:2:4, :]

ArrayInterface.size(u)  # returns (4, 6) as expected
ArrayInterface.size(v)  # returns (3, 6) as expected
ArrayInterface.size(w)  # returns (6, ) -- the first dimension is lost
```

The issue seems to come from the `_size` function in `stridelayout.jl`, where possible range types are limited to `Base.Slice` and `AbstractUnitRange`:

```julia
@generated function _size(A::Tuple{Vararg{Any,N}}, inds::I, l::L) where {N, I<:Tuple, L}
    t = Expr(:tuple)
    for n in 1:N
        if (I.parameters[n] <: Base.Slice)
            push!(t.args, :(@inbounds(_try_static(A[$n], l[$n]))))
        elseif I.parameters[n] <: AbstractUnitRange
            push!(t.args, Expr(:ref, :l, n))
        end
    end
    Expr(:block, Expr(:meta, :inline), t)
end
```

One possible fix is to replace `AbstractUnitRange` by the more general `OrdinalRange`. Is this an acceptable solution?

jipolanco:
This PR includes two commits that are meant to improve the handling of `SubArray`s, notably those that include non-unit steps.

The first commit fixes `ArrayInterface.size` for certain `SubArray`s, and should close #85.

The second commit changes the definition of `contiguous_axis` for views with non-unit steps. For instance:

```julia
A = zeros(3,4,5);
D1 = view(A, 1:2:3, :, :)  # first dimension is discontiguous
D2 = view(A, :, 2:2:4, :)  # first dimension is contiguous

contiguous_axis(D1)  # now returns Contiguous(-1)
contiguous_axis(D2)  # now returns Contiguous(1)
```

These two examples previously returned `nothing`.

### DIFF
--- a/src/stridelayout.jl
+++ b/src/stridelayout.jl
@@ -39,18 +39,19 @@ _contiguous_axis(::Any, ::Nothing) = nothing
     n = 0
     new_contig = contig = C
     for np in 1:NP
-        if I.parameters[np] <: AbstractUnitRange
+        p = I.parameters[np]
+        if p <: OrdinalRange
             n += 1
             if np == contig
-                new_contig = n
+                new_contig = (p <: AbstractUnitRange) ? n : -1
             end
-        else
+        elseif p <: Integer
             if np == contig
                 new_contig = -1
             end
         end
     end
-    # If n != N, then an axis was indexed by something other than an integer or `AbstractUnitRange`, so we return `nothing`.
+    # If n != N, then an axis was indexed by something other than an integer or `OrdinalRange`, so we return `nothing`.
     n == N || return nothing
     Expr(:call, Expr(:curly, :Contiguous, new_contig))
 end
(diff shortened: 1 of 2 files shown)


chriselrod:
Great!

timholy:
Worth noting that unlike most other array infrastructures, we don't even require regularly-spaced steps:
```julia
julia> D1 = view(A, :, :, [1, 3, 4]);

julia> ArrayInterface.contiguous_axis(D1) === nothing
true
```

PR #90, opened 2020-11-29
More changes to support VectorizationBase/LoopVectorization

chriselrod:
These are to support running the old LoopVectorization on the new VectorizationBase (the new VectorizationBase supports Julia 1.6, the old one does not).
I'm trying to get that done within the next day or two, so that LoopVectorization will finally be usable on 1.6 and people depending on it can actually test their code.

### DIFF
--- a/src/dimensions.jl
+++ b/src/dimensions.jl
@@ -30,10 +30,20 @@ Return the names of the dimensions for `x`.
     end
 end
 @inline function dimnames(::Type{T}) where {T<:Union{Transpose,Adjoint}}
-    return _transpose_dimnames(dimnames(parent_type(T)))
+    return _transpose_dimnames(Val(dimnames(parent_type(T))))
 end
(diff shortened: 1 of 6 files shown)


src/dimensions.jl
```diff
@@ -30,10 +30,20 @@ Return the names of the dimensions for `x`.
...
+# inserting the Val here seems to help inferability; I got a test failure without it.
+function _transpose_dimnames(::Val{S}) where {S}
+    if length(S) == 1
+        (:_, first(S))
+    elseif length(S) == 2
+        (last(S), first(S))
+    else
+        throw("Can't transpose $S of dim $(length(S)).")
+    end
+end
```
chriselrod, line 44: @Tokazama You may want to look at this. I got a test failure on my local master on `@inferred(val_dimnames(y'))` before making this change.

PR #93, opened 2020-12-02
Easier use of ArrayInterface.size/axes/to_dims

Tokazama:
I realized that the way I implemented these methods was kind of obnoxious for other packages to use. Now only the following need to be defined:

```julia
Base.size(x::MyArray) = some_method_that_computes_size(x)
Base.size(x::MyArray, d) = ArrayInterface.size(x, d)
```

I also added some more tests/methods for `to_dims` to make it a bit easier to work with in different situations.


chriselrod:
Looks good to me. Ready for merge and release?

Tokazama:
Yes

PR #98, opened 2020-12-17
Make working with axes (and their types) easy

Tokazama:
While working on #94 I realized that requiring construction of strides/offsets/etc in order to get type information was really annoying sometimes. This helps mediate that issue (mostly with `axes_type`). A nice benefit of this is that a lot of arrays can just define `axes_type` instead of individually defining `known_size`, `known_offset`, and `known_stride`.

I also did some trivial clean up documentation in "stridelayout.jl"


Tokazama:
@chriselrod this is ready for review

Issue #96, opened 2020-12-09
Prettier printing for `OptionallyStaticUnitRange`?


chriselrod:
```julia
julia> using PaddedMatrices

julia> @StrideArray rand(4,5)
4×5 StrideMatrix{Tuple{StaticInt{4}, StaticInt{5}}, (true, true), Float64, 1, 0, (1, 2), Tuple{StaticInt{8}, StaticInt{32}}, Tuple{StaticInt{1}, StaticInt{1}}, PaddedMatrices.MemoryBuffer{20, Float64}} with indices StaticInt{1}():StaticInt{1}():StaticInt{4}()×StaticInt{1}():StaticInt{1}():StaticInt{5}():
 0.301859  0.54085   0.222296  0.599135   0.767651
 0.336525  0.521614  0.120895  0.891117   0.58328
 0.469067  0.580745  0.745368  0.0832662  0.0853315
 0.489557  0.187212  0.228955  0.261394   0.632375

julia> StrideArray(rand(4,5))
4×5 StrideMatrix{Tuple{Int64, Int64}, (true, true), Float64, 1, 0, (1, 2), Tuple{StaticInt{8}, Int64}, Tuple{StaticInt{1}, StaticInt{1}}, Matrix{Float64}} with indices StaticInt{1}():StaticInt{1}():4×StaticInt{1}():StaticInt{1}():5:
 0.892622  0.862319   0.130142  0.308712   0.196662
 0.600617  0.0845287  0.752346  0.612079   0.534358
 0.545485  0.329272   0.305997  0.922645   0.650952
 0.618598  0.991797   0.60527   0.0707488  0.247227
```
The
```julia
with indices StaticInt{1}():StaticInt{1}():StaticInt{4}()×StaticInt{1}():StaticInt{1}():StaticInt{5}()
with indices StaticInt{1}():StaticInt{1}():4×StaticInt{1}():StaticInt{1}():5
```
Could be a lot prettier. Of course, I should probably make the rest of the `StrideArray` type print prettier, too.

Tokazama:
Do you think it should be printed with normal integers or have a concise symbol standing in for `StaticInt`?

chriselrod:
I think `StaticInt{1}()`s for the step could be omitted, so that we have
```julia
StaticInt{1}():StaticInt{4}()×StaticInt{1}():StaticInt{5}()
```
instead of
```julia
[...]

PR #101, opened 2020-12-21
Prettier StaticInt print

Tokazama:
Changes printing as discussed in #96:

```julia
julia> StaticInt(1)
Static(1)

julia> StaticInt(1):StaticInt(10)
Static(1):Static(10)

julia> StaticInt(1):StaticInt(2):StaticInt(10)
Static(1):Static(2):Static(9)
```

The nice thing about using `Static` is that it could become a standard way of informing the user that whatever is inside the brackets is a static type, instead of every new static type having it's own print out. The downside is that you can't just copy and paste things from REPL (e.g., instead printing `StaticInt(1):StaticInt(10)`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterface"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "2.14.7"
+version = "2.14.8"
 
 [deps]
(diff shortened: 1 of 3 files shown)


chriselrod (approved):
Looks good -- ready to merge?

Tokazama:
> Looks good -- ready to merge?

Yeah, I figured testing this would be better managed if/when we have doctests

PR #107, opened 2021-01-18
Better test coverage for ranges

Tokazama:


chriselrod:
More tests the merrier, thanks!

PR #114, opened 2021-02-02
Get tests to pass when running Julia with `--compiled-modules=no`.

chriselrod:
This fixes https://github.com/chriselrod/LoopVectorization.jl/issues/192#issuecomment-771332647 for me: Trixi's tests pass, as do ArrayInterface's after starting Julia with `--compiled-modules=no`.

Normally, all functions in a module have the same world age, so the order in which generated functions are defined with respect to everything else is irrelevant. But not with `--compiled-modules=no`, which causes world age errors whenever an `@generated` function calls a method defined later.

I addressed this the lazy way, but just searching for all the `@generated` functions calling non-base methods and sticking them in a file at the end. 

We could certainly do a lot better organizationally.

@Tokazama I'm not sure why it isn't letting me request you as a reviewer.


Tokazama:
> We could certainly do a lot better organizationally

Agreed. Did you have any suggestions? I've been trying to move away from having one gigantic file but I probably complicated things in the process

chriselrod:
> Agreed. Did you have any suggestions? I've been trying to move away from having one gigantic file but I probably complicated things in the process

I think breaking it up into indexing/dimensions/etc is fine, but breaking things up into "is it a `@generated` function", which this PR did, is not. Except that it fixes world age problems introduced by using `--compiled-modules=no`.

However, I now deleted `generatedfuncs.jl`, moving most of the code back, with a much more minor reordering to make sure we don't get world age issues under `--compiled-modules=no`. `@generated` definitions should all be following the definitions they're using now, or at least all of them covered by our unit tests.

So this PR is in much better shape/much less objectionable now. I'll merge and make a new release fairly soon.

PR #119, opened 2021-02-07
Clean `@pure` issues in "dimensions.jl" and "static.jl"

### LINKED ISSUE #115: Using `Base.@pure` in ArrayInterface causes 265 problems
https://github.com/JuliaLang/julia/issues/265#issuecomment-772555469

I think we should err on the side of finding any other way.

Tokazama:
This doesn't completely fix #115, but it does take care of the problems with `Symbol`. I'll update that issue with what still needs to be worked on.

Although this commit changes how `dimnames` works and completely gets rid of a couple methods it shouldn't be breaking because none of these are actually part of the README and no packages should depend on them at this point.

I also moved `static` to the README because I think it's worth making part of the public API. I think it' more convenient than typing out the entirety of a static types name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ Returns the rank of each stride.
 
 ## is_column_major(A)
 
-Returns a `Val{true}()` if `A` is column major, and a `Val{false}()` otherwise.`
+Returns a `Val{true}()` if `A` is column major, and a `Val{false}()` otherwise.
 
 ## dense_dims(::Type{T})
 Returns a tuple of indicators for whether each axis is dense.
@@ -207,6 +207,10 @@ For example, if `A isa Base.Matrix`, `offsets(A) === (StaticInt(1), StaticInt(1)
 ## can_avx(f)
 
 Is the function `f` whitelisted for `LoopVectorization.@avx`?
+
+## static(x)
+Returns a static form of `x`. If `x` is already in a static form then `x` is returned. If
+there is no static alternative for `x` then an error is thrown.
 
 ## StaticInt(N::Int)
 
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ If unknown, it returns `nothing`.
 
 ## contiguous_axis_indicator(::Type{T})
 
-Returns a tuple of boolean `Val`s indicating whether that axis is contiguous.
+Returns a tuple of boolean `StaticBool`s indicating whether that axis is contiguous.
 
 ## contiguous_batch_size(::Type{T})
 
@@ -167,7 +167,7 @@ Returns the rank of each stride.
 
 ## is_column_major(A)
 
-Returns a `Val{true}()` if `A` is column major, and a `Val{false}()` otherwise.
+Returns a `True` if `A` is column major, and a `True/False` otherwise.
 
 ## dense_dims(::Type{T})
 Returns a tuple of indicators for whether each axis is dense.
--- a/src/static.jl
+++ b/src/static.jl
@@ -260,7 +260,7 @@ Base.any(::Tuple{Vararg{True}}) = true
 Base.any(::Tuple{Vararg{Union{True,False}}}) = true
 Base.any(::Tuple{Vararg{False}}) = false
 
-@inline nstatic(::Val{N}) where {N} = ntuple(i -> StaticInt(i), Val(N))
+@inline nstatic(::Val{N}) where {N} = ntuple(StaticInt, Val(N))
 
 invariant_permutation(::Any, ::Any) = False()
 function invariant_permutation(x::T, y::T) where {N,T<:Tuple{Vararg{StaticInt,N}}}
@@ -421,6 +421,11 @@ static(:x)
 
 ```
 """
+function static end
+@aggressive_constprop static(x::Int) = StaticInt(x)
+@aggressive_constprop static(x::Bool) = StaticBool(x)
+@aggressive_constprop static(x::Symbol) = StaticSymbol(x)
+@aggressive_constprop static(x::Tuple{Vararg{Any}}) = map(static, x)
 function static(x)
     if is_static(x) isa True
         return x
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -208,6 +208,10 @@ For example, if `A isa Base.Matrix`, `offsets(A) === (StaticInt(1), StaticInt(1)
 
 Is the function `f` whitelisted for `LoopVectorization.@avx`?
 
+## static(x)
```
chriselrod, line 211: Are we at the point yet where we should just create a `Static.jl` and depend on it?

We have static `Int`s, `Bool`s, and now `Symbol`s.
Octavian also has [staticfloats](https://github.com/JuliaLinearAlgebra/Octavian.jl/blob/master/src/staticfloats.jl).

README.md
```diff
@@ -167,7 +167,7 @@ Returns the rank of each stride.
 
 ## is_column_major(A)
 
-Returns a `Val{true}()` if `A` is column major, and a `Val{false}()` otherwise.`
+Returns a `Val{true}()` if `A` is column major, and a `Val{false}()` otherwise.
```
chriselrod, line 170: Should probably have made these return `True()`/`False()`. Should have caught that before 3.0.

src/static.jl
```diff
@@ -260,25 +260,27 @@ Base.any(::Tuple{Vararg{True}}) = true
 Base.any(::Tuple{Vararg{Union{True,False}}}) = true
 Base.any(::Tuple{Vararg{False}}) = false
 
-Base.@pure nstatic(::Val{N}) where {N} = ntuple(i -> StaticInt(i), Val(N))
+@inline nstatic(::Val{N}) where {N} = ntuple(i -> StaticInt(i), Val(N))
```
chriselrod, line 263: Why not
```julia
@inline nstatic(::Val{N}) where {N} = ntuple(StaticInt, Val(N))
```
Removing an anonymous function probably makes things ever so slightly easier on the compiler. Slightly fewer allocations (actual times are very noisy):
```julia
> julia -O3 --startup=no -e "@time begin; foo(::Val{N}) where {N} = ntuple(i -> Val(i), Val{N}()); foo(Val(8)); end"                                                                 (base)
  0.013163 seconds (14.18 k allocations: 1.021 MiB, 114.01% compilation time)
> julia -O3 --startup=no -e "@time begin; foo(::Val{N}) where {N} = ntuple(Val, Val{N}()); foo(Val(8)); end"                                                                         (base)
  0.013611 seconds (11.99 k allocations: 836.604 KiB, 113.98% compilation time)
```

src/static.jl
```diff
@@ -366,11 +376,82 @@ IfElse.ifelse(::True, x, y) = x
...
+        return x
+    else
+        _no_static_type(x)
+    end
+end
 static(x::Int) = StaticInt(x)
 static(x::Bool) = StaticBool(x)
+static(x::Symbol) = StaticSymbol(x)
+static(x::Tuple{Vararg{Any}}) = map(static, x)
+@generated static(::Val{V}) where {V} = :($(static(V)))
```
chriselrod, line 435: Why not just?
```julia
@generated static(::Val{V}) where {V} = static(V)
```

src/static.jl
```diff
@@ -366,11 +376,82 @@ IfElse.ifelse(::True, x, y) = x
...
+    if is_static(x) isa True
+        return x
+    else
+        _no_static_type(x)
+    end
+end
 static(x::Int) = StaticInt(x)
 static(x::Bool) = StaticBool(x)
+static(x::Symbol) = StaticSymbol(x)
+static(x::Tuple{Vararg{Any}}) = map(static, x)
```
chriselrod, line 434: Perhaps we should try the aggressive constprop macro here, just for the sake of it?

I don't know the exact version number that added it, but if you want a safe bound:
```julia
help?> Base.@aggressive_constprop
  @aggressive_constprop ex
  @aggressive_constprop(ex)


  @aggressive_constprop requests more aggressive interprocedural constant propagation for the annotated function. For a method where the return type depends on the value of the arguments, this can yield improved inference results at the cost of additional compile time.

julia> VERSION
v"1.7.0-DEV.421"
```
So, for example (`@static` necessary to avoid `ERROR: LoadError: UndefVarError: @aggressive_constprop not defined`)
```julia
@static if VERSION >= v"1.7.0-DEV.421"
    Base.@aggressive_constprop static(x::Int) = StaticInt(x)
    Base.@aggressive_constprop static(x::Bool) = StaticBool(x)
    Base.@aggressive_constprop static(x::Symbol) = StaticSymbol(x)
    Base.@aggressive_constprop static(x::Tuple{Vararg{Any}}) = map(static, x)
else
    static(x::Int) = StaticInt(x)
    static(x::Bool) = StaticBool(x)
    static(x::Symbol) = StaticSymbol(x)
    static(x::Tuple{Vararg{Any}}) = map(static, x)
end
```
And similar everywhere else that the return type depends on values.

chriselrod (approved):
Looks good. My primary suggestion is to add `Base.@aggressive_constprop`, either using `@static if VERSION >= v"1.7.0-DEV.421"` (or whatever the specific version that added it is, if you want to spend the time to find out) to switch between two definitions, or maybe make one definition of each function annotated with `@aggressive_constprop` and have:
```julia
@static if VERSION >= v"1.7.0-DEV.421"
    using Base: @aggressive_constprop
else
    macro aggressive_constprop(ex)
        ex
    end
end
```

README.md
Tokazama: I was thinking the same thing. There is already StaticNumbers.jl but I think there is a lot of utility in a _very_ restricted set of static types. I think there has been a tendency with projects that deal with "staticness" to just solve everything with a new static type, which creates another problem.

src/static.jl
Tokazama: Dang! That is just an obvious improvement on what I was doing.

README.md
Tokazama: Where do think a package like `Static.jl` should go?

chriselrod: I just created [SciML/Static.jl](https://github.com/SciML/Static.jl) and added you as a contributor.

I'm not sure about ArrayInterface's functionality vs StaticNumbers.jl, but ArrayInterface's has definitely been growing.
I should've added static ranges to the list of things we support here.

But I do agree on the idea of having a narrower scope / focusing on being a lightweight dependency that other packages that want such functionality (including of course ArrayInterface.jl) can cheaply depend upon.

chriselrod:
Looks good. Merge when ready.

PR #122, opened 2021-02-17
Add stridelayout support for `BitArray`

chriselrod:
While the stridelayout interface is required for `device` to return `CPUPointer`, it can still be implemented for types returning `CPUindex`.
In the case of `BitArray`s, `device` should probably be returning `CPUIndex`, as it doesn't work like ordinary arrays.
VectorizationBase and LoopVectorization provide special handling for `BitArray`s to make them work in many circumstances, but I can reduce the amount of special handling required via supporting the `stridelayout` API, which requires only a handful of definitions.
The support has to be added here to avoid type piracy.


chriselrod:
Tests passed on master locally, but my master was several days old. Rebuilding to test on the latest.

chriselrod:
I get the test failure on the latest master.
I'd have to dig deeper to find out what's causing the error. Aqua is throwing the error inside `test_ambiguities`, but it does still work for other libraries I tried it on in master.

chriselrod:
I've added `AbstractRange`s as well.

Re the test failures on nightly:
https://github.com/JuliaLang/julia/issues/39698

chriselrod:
Those changes made `is_column_major` return `::True`, so I updated the tests accordingly.
But maybe we should hard-code it to `False`, since `column_major` is (a) about arrays of dimension >= 2, and (b) an explicit statement about the underlying memory.
Neither is applicable to `AbstractRange`s. Thoughts?

Tokazama:
The only problem I can think of with returning `True` is that if you wanted to have the same memory layout for two arrays and one was row major, then this would make you think `AbtractRange` is not compatible.

chriselrod:
But the range is 1-dimensional. Whatever analysis should consider that.

Tokazama:
Good point. Probably better to just be accurate then and return `False`.

chriselrod:
I meant that because column and row major vectors are equivalent, and ranges are vectors, but I'll make it `False()`.

Tokazama (commented):
LGTM

PR #123, opened 2021-02-17
Fix `device`

Tokazama:
Fixing `device` following discussion in #120. Default is now `CPUIndex` and no default for `StridedArray`. Instead, array types that work with `pointer` and define strides should explicitly define this

```julia
ArrayInterface.device(::Type{T}) where {T<:NewType} = device(parent_type(T))
```

This should also makes it so GPUArrays won't need to define `device` for all the `StridedArray` types because the proper `AbstractDevice` will be pulled from the original parent type if strides are defined.

### DIFF
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -604,7 +604,7 @@ struct CPUIndex <: AbstractCPU end
 struct GPU <: AbstractDevice end
 
 """
-device(::Type{T})
+    device(::Type{T})
 
 Indicates the most efficient way to access elements from the collection in low-level code.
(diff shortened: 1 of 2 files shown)


chriselrod:
Once upon a time I thought that `StrideArray`'s purpose was to say it supports BLAS, which requires both strided memory and `pointer` (or `Base.unsafe_convert`) to work.

Tokazama:
I'm still trying to figure out the ambiguity on nightly, but does this approach look correct?

chriselrod:
> I'm still trying to figure out the ambiguity on nightly

https://github.com/JuliaLang/julia/issues/39698

PR #124, opened 2021-02-18
Prep for new release

Tokazama:
Mostly moving code from static.jl to Static package


Tokazama:
@chriselrod , let me know this works for you and I'll get a new release out.

chriselrod:
I'll test the latest LoopVectorization release with ArrayInterface master real quick.

chriselrod:
VectorizationBase's tests fail with:
```julia
st
      Status `~/Documents/progwork/julia/env/arrayinterfaceintegration/Project.toml`
  [4fba245c] ArrayInterface v3.1.1 `~/.julia/dev/ArrayInterface`
  [bdcacae8] LoopVectorization v0.11.2
  [3d5dd08c] VectorizationBase v0.18.10
```
but pass with the latest release.
The failures all look like:
```julia
  Got exception outside of a @test
  MethodError: no method matching strides(::VectorizationBase.StridedPointer{Float64, 2, 1, 0, (1, 2), Tuple{StaticInt{8}, Int64}, Tuple{StaticInt{1}, StaticInt{1}}})
  You may have intended to import Base.strides
  Closest candidates are:
    strides(::Any) at /home/chriselrod/.julia/dev/ArrayInterface/src/stridelayout.jl:419
    strides(::A, ::Integer) where A at /home/chriselrod/.julia/dev/ArrayInterface/src/stridelayout.jl:462
    strides(::Any, ::Any) at /home/chriselrod/.julia/dev/ArrayInterface/src/stridelayout.jl:461
    ...
  Stacktrace:
    [1] _stride_error(x::Any)
      @ ArrayInterface ~/.julia/dev/ArrayInterface/src/stridelayout.jl:393
    [2] strides(x::VectorizationBase.StridedPointer{Float64, 2, 1, 0, (1, 2), Tuple{StaticInt{8}, Int64}, Tuple{StaticInt{1}, StaticInt{1}}})
      @ ArrayInterface ~/.julia/dev/ArrayInterface/src/stridelayout.jl:420
```
but they pass with the latest release.

The problem is that it uses `ArrayInterface.strides` internally, but only defined a `Base.strides` method, assuming that `ArrayInterface.strides` will fall back on `Base.strides`.

chriselrod:
Why not change [this line](https://github.com/SciML/ArrayInterface.jl/blob/bc560cf29e9aa1362852a2244c3b6dd7d57d974a/src/stridelayout.jl#L420):
```julia
defines_strides(x) || _stride_error(x)
```
to
```julia
defines_strides(x) || return Base.strides(x)
```
and let `Base.strides` throw the method error?

Alternatively, I could make a new release where I explicitly define both `ArrayInterface.strides` and `Base.strides`.

Tokazama:
> `defines_strides(x) || return Base.strides(x)`

I think this is fine. Then `defines_strides` is an opt in for `ArrayInterface.strides` rather than throwing an error.

Tokazama:
@chriselrod , did the last change for `strides` fix things for you?

chriselrod:
VectorizationBase was using functions like `ne`.

chriselrod:
VectorizationBase's tests pass, but LoopVectorization got a failure because of `device` now returning `CPUIndex`.
https://github.com/SciML/ArrayInterface.jl/pull/123
But that's because it defined the following array:
```julia
struct SizedOffsetMatrix{T,LR,UR,LC,UC} <: DenseMatrix{T}
    data::Matrix{T}
end
Base.size(::SizedOffsetMatrix{<:Any,LR,UR,LC,UC}) where {LR,UR,LC,UC} = (UR-LR+1,UC-LC+1)
Base.axes(::SizedOffsetMatrix{T,LR,UR,LC,UC}) where {T,LR,UR,LC,UC} = (Static{LR}():Static{UR}(),Static{LC}():Static{UC}())
Base.parent(A::SizedOffsetMatrix) = A.data
Base.unsafe_convert(::Type{Ptr{T}}, A::SizedOffsetMatrix{T}) where {T} = pointer(A.data)
ArrayInterface.contiguous_axis(::Type{<:SizedOffsetMatrix}) = ArrayInterface.One()
ArrayInterface.contiguous_batch_size(::Type{<:SizedOffsetMatrix}) = ArrayInterface.Zero()
ArrayInterface.stride_rank(::Type{<:SizedOffsetMatrix}) = (ArrayInterface.StaticInt(1), ArrayInterface.StaticInt(2))
function ArrayInterface.strides(A::SizedOffsetMatrix{T,LR,UR,LC,UC}) where {T,LR,UR,LC,UC}
    (Static{1}(), (Static{UR}() - Static{LR}() + Static{1}()))
end
ArrayInterface.offsets(A::SizedOffsetMatrix{T,LR,UR,LC,UC}) where {T,LR,UR,LC,UC} = (Static{LR}(), Static{LC}())
Base.getindex(A::SizedOffsetMatrix, i, j) = LoopVectorization.vload(LoopVectorization.stridedpointer(A), (i,j))
```

While `LoopVectorization.check_args(skern)` now returns `false` (when it returned `true` earlier), which protects it from `@avx`, `getindex` was explicitly defined to use `stridedpointer`, so the fallback of using regular indexing throws an error.

But I guess that's fine. There probably isn't any code doing this "in the wild".
`StrideArrays`/`StrideArraysCore` were already explicitly defining `device` to be `CPUPointer`, so they're okay.

Tokazama:
This is a tough one. I keep going back and forth on how much gait keeping we should do with this. From the `DenseArray` docstring it seems like an array like `SArray` could be a subtype, but that obviously would be `CPUIndex`. Theoretically, if we knew the strides of an array and how to get to the parent memory we could guarantee this is safe, but the fallback for `pointer(::AbstractArray)` [doesn't give any guarantees](https://github.com/JuliaLang/julia/blob/3129a5bef56bb7216024ae606c02b413b00990e3/base/abstractarray.jl#L1116) that the pointer will be valid. 

Maybe this?

```julia

function device(::Type{T}) where {T<:AbstractArray2}
    if (parent_type(T) <: T) && !defines_strides(T)
        return CPUIndex()
    else
        return device(parent_type(T))
    end
end

function Base.pointer(a::A) where {A<:AbstractArray2}
    if parent_type(A) <: A
        return unsafe_convert(Ptr{T}, x)
    else
        return pointer(parent(a))
    end
end
```

That way inheriting memory access patterns for a subtype of `AbstractArray2` only requires ensuring you don't change strides and doing `ArrayInterface.defines_strides(::Type{T}) where {T<:NewType} = true`.

chriselrod:
>  but the fallback for `pointer(::AbstractArray)` [doesn't give any guarantees](https://github.com/JuliaLang/julia/blob/3129a5bef56bb7216024ae606c02b413b00990e3/base/abstractarray.jl#L1116) that the pointer will be valid.

But the [strided array interface in base](https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-strided-arrays) requires that convert method.

Because `DenseArray`s are part of the `StridedArray` union, `StaticArray`s should not be dense.
I do think this is enough of a reason to make strided arrays `CPUPointer` (and that all GPUArrays are not Dense arrays/their subtyping is invalid).

Tokazama:
That makes sense. So maybe we have

```julia
function device(::Type{T}) where {T}
    if (parent_type(T) <: T) && !defines_strides(T)
        return CPUIndex()
    else
        return device(parent_type(T))
    end
end
```

And explicitly mention in the `defines_strides` docs that `pointer(x::T)` should work?

chriselrod:
> That makes sense. So maybe we have
> 
> ```julia
> function device(::Type{T}) where {T}
>     if (parent_type(T) <: T) && !defines_strides(T)
>         return CPUIndex()
>     else
>         return device(parent_type(T))
>     end
> end
> ```
> 
> And explicitly mention in the `defines_strides` docs that `pointer(x::T)` should work?

This would infinitely recurse if `parent_type(T) <: T && defines_strides(T)`, so maybe
```julia
function device(::Type{T}) where {T}
    if (parent_type(T) <: T)
        defines_strides(T) ? CPUPointer() : CPUIndex()
    else
        return device(parent_type(T))
    end
end
```
?

Tokazama:
Oh good point. Totally missed that...

chriselrod:
What happened to "Fall back to Base.strides when define_strides(x) == false"?
Seems to have been reverted.

Tokazama:
Sorry, I thought that issue was related to subtypes of `DenseArray` and the more recent fix was an alternative.

chriselrod:
VectorizationBase was defining `Base.strides` and calling `ArrayInterface.strides`. I'm backporting defining `ArrayInterface.strides`:
https://github.com/chriselrod/VectorizationBase.jl/commit/77564fa3347ad21a0fe3ff0a36cbec3b709a8699

Tokazama:
I just want to make sure I understand correctly. Should I not include the `Base.strides` fall back now?

chriselrod:
I don't see why not to include it, but not including it at least shouldn't break things now.

chriselrod:
VectorizationBase's and LoopVectorization's tests passed locally on this PR.

chriselrod (approved):
Given that my tests pass, shall we get this out the door?

PR #131, opened 2021-03-05
Add CPUTuple <: AbstractCPU as new device type

chriselrod:
Related to https://github.com/JuliaArrays/ArrayInterface.jl/issues/130

I'm open to suggestions on a better name, e.g., if someone were to define:
```julia
struct My3Vector{T} <: AbstractVector{T}
    x::T
    y::T
    z::T
end
```
it should also be a `CPUTuple`. So maybe something like `CPUStruct` is a better name? Or `CPUStructMemory`?
Or `CPULLVMArray`? Most homogenous tuples lower to LLVM Arrays.

The idea is to represent something that should really be `CPUPointer`, except for the niggling detail that Julia semantics don't allow us to get a pointer to it, while C(++) would allow us to just use the `&` operator to get the address.

Julia's semantics actually more closely match LLVM there, and what Clang does when you `&` is basically the same as calling `Ref` in Julia and then using that `Ptr`, except maybe Clang optimizes away the copy more consistently for some reason? That'll take some more exploration, I just recall in my tests some time ago Julia often failed to optimize them away, but I didn't actually check whether a similar example in C succeeds.

So the `CPUTuple` type means to say that this is what we should do.

That's to distinguish it from other `AbstractArray` representations that aren't actually backed by memory underlying their loads, like `Fill` arrays, ranges, etc.


Tokazama:
So I know that you can't go from `Tuple` to a pointer, but if we had something like this...

```julia
buffer = Ref{NTuple{N,T}}()
a = SArray{S,T,N,L}(buffer[])
```
...does the dereferencing always reallocate that entire buffer to create `a`?

chriselrod:
> ...does the dereferencing always reallocate that entire buffer to create `a`?

No, sometimes LLVM is in a good mood.

Tokazama:
So is the issue that we want to be able to do something like...
```
dereference(::CPUStruct, buffer) = buffer[] # get rid of reference and mutability
dereference(::CPUPointer, buffer) = buffer  # keep reference and mutability, other functions wrap in array struct
```

chriselrod:
> So is the issue that we want to be able to do something like...
> 
> ```
> dereference(::CPUStruct, buffer) = buffer[] # get rid of reference and mutability
> dereference(::CPUPointer, buffer) = buffer  # keep reference and mutability, other functions wrap in array struct
> ```

I'll redefine [VectorizationBase.memory_reference](https://github.com/chriselrod/VectorizationBase.jl/blob/088edc8e65f8ea8f39a0878bc66fd456f0123caa/src/strided_pointers/stridedpointers.jl#L35) to return a tuple containing "ptr" and memory, to something like
```julia
@inline memory_reference(A::BitArray) = (Base.unsafe_convert(Ptr{Bit}, A.chunks), A.chunks)
@inline memory_reference(A::AbstractArray) = memory_reference(device(A), A)
@inline memory_reference(::CPUPointer, A) = (pointer(A), preserve_buffer(A))
@inline function memory_reference(::CPUTuple, A::AbstractArray{T}) where {T}
    ref = Ref(A)
    Base.unsafe_convert(Ptr{T}, ref), ref
end
```
Then the suggested use would be something like:
```julia
# either
ptra, presa = stridedpointer_and_buffer(A);
ptrb, presb = stridedpointer_and_buffer(B);
# or
(ptra,ptrb), (presa,presb) = groupedstridedpointer((A,B), (#= description of axis similarity =#));
GC.@preserve presa presb begin
    # use ptra and ptrb
end
```
This would allow `SArray`s and `NTuple`s to work, but would rely on LLVM being generous for good performance (unlike `MArray`s, which should perform well reliably).

However, the `@gc_preserve` macro would probably just pass through `CPUTuple`s, the idea being it's intended to let arrays be stack allocated, which should already be the case for `CPUTuple` objects.

I'll make a breaking change to `groupedstridedpointer` to also return the buffer, figuring that no one is actually using it yet (except for a soon-to-be-mainlined LoopVectorization branch).

Tokazama (commented):
LGTM

Issue #130, opened 2021-03-03
Array Constructors


Tokazama:
I think we are in a good place to start working on array constructors (would also make documenting examples a whole lot easier). This brings up stuff  related to`similar`:

* https://github.com/JuliaLang/julia/issues/25107
* https://github.com/JuliaLang/julia/issues/18161
* https://github.com/JuliaLang/julia/issues/10064

I think we actually need to break up what similar does into several pieces though

1. allocate memory - a buffer/Ref that is safer than a pointer
2. do something to the memory - probably fastest on pointers
3. create user facing instance - e.g., Array

I haven't worked out all the details but here's some of the cleaner code I have so far that might support this.

```julia
""" allocate_memory(x, args...) """  # step 1 allocates memory with corresponding device to `x` and axes
function allocate_memory(x)  end

""" preserve(f, x, y) """  # step 2 operate on pointers and permit novel garbage collection approaches
preserve(f, x::X) where {X} = preserve(f, x, device(X))
preserve(f, x::X, ::CPUIndex) where {X} = f(x)
function preserve(f, x::X, ::CPUPointer) where {X}
    GC.@preserve x out = f(pointer(x))
    return out
end
preserve(f, x::X, y::Y) where {X,Y} = preserve(p_x -> preserve(Base.Fix1(op, p_x), y), x)


""" as_immutable """
function as_immutable(x) end

""" initialize(original, new_data) """ # step 3 turn processed buffer into user facing array
function initialize(x::X, data) where {X}
    if !ismutable(X)
        return as_immutable(data)
    else
        return data
    end
end

""" instantiate(f, x, allocator, initiializer) """
instantiate(f, x, allocator, initiializer) = initiializer(x, preserve(f, allocator(x),  x))
instantiate(f, x, allocator) = (f, x, allocator, initiialize)
instantiate(f, x) = (f, x, allocate_memory)
```

The reason I think separating things out like this is helpful is because it turns this function from base like this
```julia
function rotr90(A::AbstractMatrix)
    ind1, ind2 = axes(A)
    B = similar(A, (ind2,ind1))
    m = first(ind1)+last(ind1)
    for i=ind1, j=axes(A,2)
        B[j,m-i] = A[i,j]
    end
    return B
end
```
into this
```julia
function rotr90(A)
    ind1, ind2 = axes(A)
    return instantiate(A, x -> allocate_memory(x, (ind2, ind1))) do a, b
        m = first(ind1)+last(ind1)
[...]

PR #133, opened 2021-03-08
Implement `broadcast_axis`

Tokazama:
This is essentially a formalization of [_bcs1](https://github.com/JuliaLang/julia/blob/86387a8587b7c3f6ecb28898d91365c2ae60a6a9/base/broadcast.jl#L495) that is called within the `combine_axes` function for broadcasting.

I'm sure we want to do some more stuff for broadcasting later, but for now I'm just trying to get some basic stuff in for combining axes. I'm not sure if we want the function that iteratively calls `broadcast_axis` to be `combine_axes`. We may want something more like `combine_layout` so that it can can account for reconstructing various types of sparsely populated arrays, which would involve more than just axes.

### DIFF
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -792,6 +792,7 @@ include("dimensions.jl")
 include("axes.jl")
 include("size.jl")
 include("stridelayout.jl")
+include("broadcast.jl")
 
 
 abstract type AbstractArray2{T,N} <: AbstractArray{T,N} end
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -32,7 +32,7 @@ broadcast_axis(::BroadcastAxisDefault, x, y) = _broadcast_axis(BroadcastAxis(y),
 function _broadcast_axis(::BroadcastAxisDefault, x, y)
     return One():_combine_length(static_length(x), static_length(y))
 end
-_broadcast_axis(s::BroadcastAxis, x, y) = broadcasted_axis(s, x, y)
+_broadcast_axis(s::BroadcastAxis, x, y) = broadcast_axis(s, x, y)
 
 # we can use a similar trick as we do with `indices` where unequal sizes error and we just
 # keep the static value. However, axes can be unequal if one of them is `1` so we have to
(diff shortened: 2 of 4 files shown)


Tokazama:
Does this look good for now? It lacks the context of the entire broadcasting system but I figured it would be more manageable to approach it in a series of small PRs.

Tokazama:
I should probably give some more context for this.

Whether the broadcasting struct (something like `Broadcasted`) is passed to `instantiate` or `materialize` the axes have to be combined in some way. Typically these are passed to similar, allowing each element produced by the broadcasted function to be computed in place.
The most obvious benefit of having something like `broadcast_axis` is it helps properties related to the axis, resolving issues likes this https://github.com/JuliaArrays/StaticArrays.jl/issues/744.

It also helps with #130, where we might not actually be broadcasting something bet we are performing some sort of operation on two arrays that needs a destination that isn't a fully materialized array.

Tokazama:
Was the consensus that I needed to address how we're going to interact with the entirety  broadcasting before moving forward with this?

chriselrod:
Sorry for being slow here. What is the interface/way for someone to use this?

Tokazama:
The big pictures is that I'm trying to get a coherent interface for breaking down arrays into their memory and structural layout so we can have generic methods that help us reconstruct arrays (like `similar` but more flexible). 

I've packaged the code I've been working on around this [here](https://github.com/Tokazama/MemoryBuffers.jl/blob/main/src/MemoryBuffers.jl) and [this line](https://github.com/Tokazama/MemoryBuffers.jl/blob/871dc0835d12461ee52cc1c1e4530d1609bfc275/src/MemoryBuffers.jl#L158) is where broadcasting stuff is implemented. You can see where things start to diverge from `Base.Broadcast` when `combine_layouts` is called instead of `combined_axes`. I didn't include any of the layout stuff in this PR because it would require figuring out buffers too and I was trying to keep this one simple. I can add it in here if it makes more since though.

> Is it something someone opts into by defining new range types (returned by axes), broadcast styles, or that a different broadcasting implementation opts into?

We could just us a different subtype of `AbstractBroadcasted` and have it diverge at `instantiate` and/or `materialize!`.

chriselrod (approved):
Okay, sounds good.

PR #137, opened 2021-04-04
General Improvements to indexing

Tokazama:
Highlights:
* Improved docs for `indices`
* consolidated the `flatten_*` code within `to_indices`
* removed superfluous keyword arguments from several functions
* Added support for `IndexCartesian` style `to_index`
* Multidim -> linear indexing

The last one gives support for statically sized arrays. Here's some output for reference. Calling `ArrayInterface.getindex` appears similar but doesn't preserve the static value.

```julia
julia> x = LinearIndices((static(0):static(3),static(3):static(5),static(-2):static(0)));

julia> @code_llvm ArrayInterface.to_index(x, (static(0), static(3), static(-2)))
;  @ /Users/zchristensen/projects/ArrayInterface.jl/src/indexing.jl:173 within `to_index'
define void @julia_to_index_283() {
top:
  ret void
}

julia> @code_typed ArrayInterface.to_index(x, (static(0), static(3), static(-2)))
CodeInfo(
1 ─     return static(1)
) => StaticInt{1}
```


src/stridelayout.jl
```diff
@@ -38,6 +38,14 @@ function _offsets(x::X, dim::StaticInt{D}) where {X,D}
     end
 end
 
+"""
+    offset1(x) -> Integer
+
+Returns the offset of the linear indices for `x`.
+"""
+offset1(x) = _offset1(has_parent(x), x)
```
chriselrod, line 46: > Regardless of the array's native indices, linear indices always range from 1:length(A). However, this raises an ambiguity for one-dimensional arrays (a.k.a., AbstractVector): does v[i] mean linear indexing , or Cartesian indexing with the array's native indices?

https://docs.julialang.org/en/v1/devdocs/offset-arrays/#Linear-indexing-(LinearIndices)

Maybe we should make the signature `offset1(::AbstractVector)`, define a generic `offset1(::AbstractArray) = static(1)`?

chriselrod: But I guess other array types don't have to be beholden to this, since it is just the `OffsetArray`s docs.

chriselrod (approved):
I didn't look too closely, but it didn't touch the stride layout code so I don't anticipate any problems on my end.

src/stridelayout.jl
Tokazama: Originally I was trying to get this to the point where we could retrieve the value from a pointer and views of arrays might have different `offset1` values. We'll probably need a more elaborate definition in the future.

Tokazama:
Then I'll go ahead and merge this. The only changes people should expect is better indexing when statically sized arrays are indexed.

PR #139, opened 2021-04-05
Fix strides for view of a StaticArray.

chriselrod:


chriselrod:
Should we change the test bounds on ArrayInterface?
Is 1.2 no longer supported for CI, or why does it always time out, seemingly without actually running any tests?

Tokazama:
> Should we change the test bounds on ArrayInterface?

I'd be fine with that. If anyone is stuck at an old version because we drop it then hopefully they'd upgrade by the next Julia LTS

chriselrod:
The latest versions of...require...
DifferentialEquations requires Julia 1.3
Images requires Julia 1.3
LoopVectorization requires Julia 1.5

So perhaps 1.3 would be a good target if we actually want to drop support.

However, Julia 1.5's tests haven't completed yet either. Not sure where it is hanging, I'll have to take a look.

Tokazama:
1.3 would be fine too. I remember having trouble specifically with 1.2 a while ago in AxisIndices.jl. I think that's when they added a bunch of threading stuff and I couldn't use it.

Tokazama (commented):
Looks like Aqua was the problem?
This looks good to me.

chriselrod:
Yeah. Turning off Aqua for <1.6 let tests pass again on 1.2, so I left the version bounds there.

PR #140, opened 2021-04-06
NDIndex - Optionally static CartesianIndex

Tokazama:
While working on more indexing stuff I needed the ability to propagate static info through something like `CartesianIndex`. I also saw there was a variant of this sort of thing in VectorizationBase.jl, so I think it could be widely beneficial to start getting something standard in place.

### DIFF
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -27,8 +27,6 @@ else
     end
 end
 
-static_ndims(x) = static(ndims(x))
-
 if VERSION ≥ v"1.6.0-DEV.1581"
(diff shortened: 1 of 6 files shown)


chriselrod (approved):
Seems fine, but I realize now that I VectorizationBase/LoopVectorization shouldn't need this anymore.
Once upon a time, ranges all round-tripped through `static_first(r):static_last(r)`, so I needed static `CartesianIndex` to preserve statically sized `CartesianIndices`.
Now it uses `LoopVectorization._canonicalize_range` instead.

Tokazama:
It's not necessarily something we absolutely can't live without, but it also gives you the nice guarantee that you have a bunch of `Int`s and/or `StaticInt`s.

PR #141, opened 2021-04-11
Layouts - combining array structure and indexing

Tokazama:
This is the first big step towards solving a lot of the recent issues I've created and also further combines the stride traits with the rest of ArrayInterface.jl.

The conceptual piece here is the concept of a "layout" (if you have a better name feel free to recommend it b/c I hate naming things). The idea is that we need to propagate information about an array's structure _and_ anything it wraps when accessing collections of its elements so I've included `layout(::AbstractArray, index)`. The return value of `layout` is intended to be the most appropriate iterator for accessing the _memory_ that corresponds to `index`. Note that this is different than `eachindex`, `LinearIndices`, or `CartesianIndices`, which are meant to directly index the array they are called on. This is why I've also included `buffer` for extracting the most direct reference to an array's memory. This PR is strictly focusing on the stride array layout, but I've played around with a bunch of different ways this could be generalized to other situations (sparse, lazy stacked). For and idea of how we are already circling this in Base, you could look at [`SCartesianIndex2`](https://github.com/JuliaLang/julia/blob/c87d85feaf547f763c5dee64c1be163b02df210d/base/reinterpretarray.jl#L202) for reinterpreted arrays.

Currently `layout` is very basic and geared towards arrays with strides. If linear indexing is performed, then we see if the array is `IndexLinear`, otherwise we construct an instance of `StrideIndices`. I've also changed what the flow into `to_indices` is. Instead of passing the indexed array the result of layout is passed. This works out pretty nicely because if you know the layout of an array you also know the axes so I've mostly done some cleaning up there.

Here are some very basic benchmarks (with the array print out removed).

```julia
julia> x = ones(4,4,4);

julia> @btime ArrayInterface.getindex(x, 3, :, :)
  74.965 ns (2 allocations: 224 bytes)

julia> @btime getindex(x, 3, :, :)
  347.114 ns (4 allocations: 288 bytes)

julia> @btime ArrayInterface.getindex(x, :, 3, :)
  69.938 ns (2 allocations: 224 bytes)

julia> @btime getindex(x, :, 3, :)
  404.116 ns (7 allocations: 352 bytes)

julia> @btime ArrayInterface.getindex(x, :, :, 3)
  73.510 ns (2 allocations: 224 bytes)

julia> @btime getindex(x, :, :, 3)
  276.487 ns (4 allocations: 288 bytes)
```

I also wanted to see how much of a difference was due to cleaning up the overall pipeline, and how much was due to using strides. So I compared comparable spots in the base call to indexing a collection and ArrayInterface.jl

```julia
inds = Base.to_indices(x, (:, :, :))
@btime inner_index(x, inds)
@btime inner_index_base(x, inds)

julia> @btime inner_index(x, inds)
  119.776 ns (1 allocation: 624 bytes)

julia> @btime inner_index_base(x, inds)
  238.631 ns (5 allocations: 688 bytes)

``` 

None of this is combining slices or taking advantage of any fancy memory management, just turning multidim indexing into linear indexing with strides.

I'll continue to put together more robust tests, but I wanted know if/how you'd like this to work with StrideArrays.jl and LoopVectorization.jl. This was discussed in some detail in #94 (which this is replacing). I was leaning towards having something like vaguely akin to...

```julia
function unsafe_get_collection(A, layout, inds)
    axs = to_axes(lyt, inds)
    dst = similar(A, axs)
    map(Base.unsafe_length, axes(dst)) == map(Base.unsafe_length, axs) || Base.throw_checksize_error(dst, axs)
    # usually a generated function, don't allow it to impact inference result
    Ab, Aptr = buffer_pointer(A)
    dstb, dstptr = buffer_pointer(A)
    GC.@preserve Ab dstb copyto!(dstptr, layout(dstptr, :), Aptr, view(lyt, inds...))
    return dst
end
```
...and then whatever buffers and pointers are produced could control dispatch for more optimized memory management based on hardware.

### DIFF
--- a/src/layouts.jl
+++ b/src/layouts.jl
@@ -1,0 +1,140 @@
+
+_as_index(x) = x
+_as_index(x::Integer) = static(1):x
+_as_index(x::OneTo) = static(1):length(x)
+_as_index(x::StepRange) = OptionallyStaticStepRange(x)
+_as_index(x::UnitRange) = OptionallyStaticUnitRange(x)
+_as_index(x::OptionallyStaticRange) = x
+
+"""
+    StrideLayout(A)
+
+Produces an array whose elements correspond to the linear buffer position of `A`'s elements.
+"""
+struct StrideLayout{N,R,O1,S,A<:Tuple{Vararg{Any,N}}} <: AbstractArray2{Int,N}
+    rank::R
+    offset1::O1
+    strides::S
+    axes::A
+end
+
+offset1(x::StrideLayout) = getfield(x, :offset1)
+offsets(x::StrideLayout) = map(static_first, axes(x))
+axes(x::StrideLayout) = getfield(x, :axes)
+@inline function axes(x::StrideLayout{N}, i::Int) where {N}
+    if i > N
+        return static(1):1
+    else
+        return getfield(getfield(x, :axes), i)
+    end
+end
+@inline function axes(x::StrideLayout{N}, ::StaticInt{i}) where {N,i}
+    if i > N
+        return static(1):static(1)
+    else
+        return getfield(getfield(x, :axes), i)
+    end
+end
+strides(x::StrideLayout) = getfield(x, :strides)
+stride_rank(x::StrideLayout) = getfield(x, :rank)
+
+@inline function StrideLayout(x::DenseArray)
+    a = axes(x)
+    return StrideLayout(
+        stride_rank(x),
+        offset1(x),
+        size_to_strides(map(static_length, a), static(1)),
+        a
+    )
+end
+
+# TODO optimize this
+@inline function StrideLayout(x)
+    return StrideLayout(
+        stride_rank(x),
+        offset1(x),
+        strides(x),
+        axes(x)
+    )
+end
+
+##############
+### layout ###
+##############
+layout(x, i) = layout(x)
+layout(x, i::AbstractVector{<:Integer}) = _maybe_linear_layout(IndexStyle(x), x)
+layout(x, i::Integer) = _maybe_linear_layout(IndexStyle(x), x)
+layout(x, i::AbstractCartesianIndex{1}) = _maybe_linear_layout(IndexStyle(x), x)
+function layout(x, i::AbstractVector{AbstractCartesianIndex{1}})
+    return _maybe_linear_layout(IndexStyle(x), x)
+end
+_maybe_linear_layout(::IndexLinear, x) = _as_index(eachindex(x))
+_maybe_linear_layout(::IndexStyle, x) = layout(x)
+layout(x::StrideLayout) = x
+layout(x::LinearIndices) = x
+layout(x::CartesianIndices) = x
+function layout(x)
+    if defines_strides(x)
+        return StrideLayout(x)
+    else
+        return _layout_indices(IndexStyle(x), axes(x))
+    end
+end
+function layout(x::Transpose)
+    if defines_strides(x)
+        return StrideLayout(x)
+    else
+        return Transpose(layout(parent(x)))
+    end
+end
+function layout(x::Adjoint{T}) where {T<:Number}
+    if defines_strides(x)
+        return StrideLayout(x)
+    else
+        return Transpose(layout(parent(x)))
+    end
+end
+function layout(x::PermutedDimsArray{T,N,perm,iperm}) where {T,N,perm,iperm}
+    if defines_strides(x)
+        return StrideLayout(x)
+    else
+        p = layout(parent(x))
+        return PermutedDimsArray{eltype(p),ndims(p),perm, iperm,typeof(p)}(p)
+    end
+end
+function layout(x::SubArray)
+    if defines_strides(x)
+        return StrideLayout(x)
+    else
+        return @inbounds(view(layout(parent(x)), x.indices...))
+    end
+end
+_layout_indices(::IndexStyle, axs) = CartesianIndices(axs)
+_layout_indices(::IndexLinear, axs) = LinearIndices(axs)
+
+"""
+    buffer(x)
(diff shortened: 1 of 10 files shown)


Tokazama:
@chriselrod , Before you review the code here I want to pass something by you:

After some exploratory work on improving this I wonder if there's a better way to do this. The approach here basically constructs something like `LinearIndices` or `CartesianIndices`, but instead of representing a single array `StrideLayout` is the unified representation of multiple arrays with strides. The end goal was that optimizing unique scoping/indexing patterns would be easier because the mapping between user provided indices and the core buffer/data would just be a
"layout". With some more cleaning up I found this improved indexing pretty consistently (still need VectorizationBase.jl to make pointers fast though).

My concern is that this approach won't translate as well with arrays that don't have strides (sparse arrays) and will be difficult to incorporate device and machine info into the codegen for VectorizationBase.jl. I started played around with something that was more based in traits but at a certain point I realized I was just imitating IR passes and probably triggering a lot of unnecessary compiler work. I took a look at the MLIR manual and came across the [Data Layout Modeling](https://mlir.llvm.org/docs/DataLayout/) section, which describes a lot of what I'm trying to accomplish here.

I haven't had time to test this out and don't know how feasible it is, but what if we had array IR here that only focused on combining nested array representations and composed the final function that iteratively copied indices to the new array? Then packages from JuliaSIMD would be able to alter this IR with info about hardware. This would change the current indexing pipeline to basically be.

1. `getindex` - index entry point
2. `to_indices` - make sure user indexing is inbounds and in the appropriate form
3. `unsafe_getindex` - generate destination array
4. `unsafe_getindex!` - composes the code the iteratively copies to the destination array.

If this worked then theoretically other array types would be a lot simpler to incorporate and optimize using ArrayInterface.jl

You know this stuff way better than I do, so if you think this would be an unhelpful distraction from what you're already trying to do then perhaps the original direction I started in here is better. Also, if you think the IR approach is right but want me to take a back seat I can just focus my efforts here on documentation and cleaning up what I've already implemented.

src/layouts.jl
```diff
@@ -0,0 +1,139 @@
...
+    StrideLayout(A)
+
+Produces an array whose elements correspond to the linear buffer position of `A`'s elements.
+"""
+struct StrideLayout{N,R,O1,S,A<:Tuple{Vararg{Any,N}}} <: AbstractArray2{Int,N}
+    rank::R
+    offset1::O1
+    strides::S
+    axes::A
+end
```
chriselrod, line 18: On the VectorizationBase side of things, it'd be nice if `StridedPointer` could be a `StrideLayout` + a `Ptr`.
However, `StrideLayout` is missing some things, most importantly the contiguous axis, and also has `axes` which it does not want (although it does want to store `map(static_first, axes(x))`).

Also, `StridedPointer` + `axes` == `PtrArray`. `StrideLayout` already subtypes `AbstractArray2`, so it is almost there itself.

src/layouts.jl
```diff
@@ -0,0 +1,139 @@
...
+function layout(x::SubArray)
+    if defines_strides(x)
+        return StrideLayout(x)
+    else
+        return @inbounds(view(layout(parent(x)), x.indices...))
+    end
+end
+_layout_indices(::IndexStyle, axs) = CartesianIndices(axs)
+_layout_indices(::IndexLinear, axs) = LinearIndices(axs)
```
chriselrod, line 112: Whether to use linear or Cartesian indexing is of course context dependent (in the rewrite, I'll add support for switching between representations).
E.g., if you're summing a `Matrix{Float64}` or a `Adjoint{Float64,Matrix{Float64}}`, you'd want to use linear indexing in both cases (for the `Adjoint`, the linear indices would be traversing it in row major order). Same if you're calculating the sum of 2 instances of `Matrix{Float64}` or 2 instances of `Adjoint{Float64,Matrix{Float64}}`. However, if you sum add a `Matrix{Float64}` to a `Adjoint{Float64,Matrix{Float64}}`, you suddenly need cartesian indexing.
`eachindex` gets some of this right, but it'll always return `IndexCartesian()` for the `Adjoint`, meaning performance for those sums/additions is worse than necessary.

Is there a "combine layouts"?

I'll get LoopVectorization to be able to transform between these correctly in the rewrite.
One of my goals there is for something like this to work:
```julia
@avx begin
  C .= 0
  for n in indices((C,B),2), m in indices((C,A),1), for k in indices((A,B),(2,1))
    C[m,n] += A[m,k] * B[k,n]
  end
end
```
and to have LoopVectorization be capable of converting the linear indexing of `C .= 0` into cartesian indexing, and then fuse that broadcast with the following loop.

chriselrod (commented):
I haven't taken a close look at the MLIR document yet, but it looks like it's talking about more low level details, of the sort covered by `Base.datatype_alignment`?

src/layouts.jl
Tokazama: What you're describing is basically what I was getting at with the reference to MLIR. The final "layout" after code gen would be the result of scoping (for summation this would be unordered access to all elements once) which informs how to combine each layer of the arrays.

src/layouts.jl
Tokazama: How about something that is indexable but isn't necessarily an iterator like this

```julia
struct StrideLayout{N,S<:Tuple{Vararg{Any,N}},R,O,O1,C}
    strides::S
    rank::R
    offset1::O1
    offsets::O
    contiguous::C
end
```

chriselrod: I don't have any actual support anywhere for `ArrayInterface.contiguous_batch_size` yet, so we could add that later, but probably best to start carrying that around now as well.
But this looks good.

Tokazama:
see #169

PR #143, opened 2021-04-26
Fix ReinterpretArray's definition of strides

chriselrod:
To be inline with ArrayInterface's meaning (i.e., about memory layout), rather than use [size_of_strides](https://github.com/JuliaLang/julia/blob/75c4f5562a0cd83f47c5695caee33c45630e76fc/base/reinterpretarray.jl#L161) like Base Julia.
Using `size_to_strides` causes LoopVectorization to get incorrect answers when `reinterpret`ing `SubArray`s.

### DIFF
--- a/src/stridelayout.jl
+++ b/src/stridelayout.jl
@@ -443,6 +443,27 @@ function strides(x)
         return Base.strides(x)
     end
 end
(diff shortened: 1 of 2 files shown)


Tokazama (commented):
LGTM

PR #149, opened 2021-05-03
Improve single element indexing (and other quality of life improvements)

Tokazama:
While working on layouts for #141 I realized that unless `getindex` was skipped (e.g., directly calling `unsafe_getindex`) than element-wise indexing was a lot slower. I approached this from a lot of different angles (some of which will probably come up in another PR still), but the issue appears to be due to the compiler not knowing to elide reconstruction of something like `Tuple{Int,Int}`, despite not undergoing any change throughout `to_indices`. So I'll start with the fix I had for that.

# Canonical Indexing Types
I chose the name because I saw a similar concept in [LoopVectorization](https://github.com/JuliaSIMD/LoopVectorization.jl/blob/2fdb50f8685c85f95e3292b137c0a8f3a9ca6c48/src/modeling/graphs.jl#L873). In `to_indices(A, args::Tuple)` each element of `args` is checked to see if it is a canonical indexing type up front, avoiding recursive reconstruction of the tuple of indexers. This check is performed by `is_canonical`, resulting in the following improvement:

```julia
x = ones(4,4,4);

#  current version
julia> @btime ArrayInterface.getindex($x, 3, 3, 3)
  4.566 ns (0 allocations: 0 bytes)
1.0

#  new version
julia> @btime ArrayInterface.getindex($x, 3, 3, 3)
  1.755 ns (0 allocations: 0 bytes)
1.0

# base version
julia> @btime getindex($x, 3, 3, 3)
  1.756 ns (0 allocations: 0 bytes)
1.0

```

I've also expanded on the number of types that are considered "canonical" indexers from that of base (e.g., we don't ever need to change `OptionallyStaticUnitRange`), so there may also be some (_very small_) improvements on indexing collection.
I think we could still squeeze some more performance out of this by avoiding construction of axes for types where the conversion to the canonical state is known already.
This is kind of the idea behind `canonicalize(x)`, which converts `x` to a canonical indexer and helped clean up some internal code that does the same thing.
I'm probably going to work on that optimization in another PR b/c it will probably require some involved work on the internal mechanics of `to_indices`.
For now we at least are as fast as Base.

# LazyAxis

A lot of array traits can be derived from looking an array's axes.
Many arrays don't have fully constructed axes stored with them, meaning we usually spend some unnecessary time composing an axis to just find out its size or offset.
Using `lazy_axes` we can sometimes avoid these costs.

For example, we can avoid the cost of retrieving size information for an `Array`.

```julia
axes_map_first(x) = map(first, ArrayInterface.axes(x))
lazy_axes_map_first(x) = map(first, ArrayInterface.lazy_axes(x))

julia> @btime axes_map_first($x)
  1.754 ns (0 allocations: 0 bytes)
(1, 1, 1)

julia> @btime lazy_axes_map_first($x)
  0.046 ns (0 allocations: 0 bytes)
(1, 1, 1)
```

Much of this improvement is lost when we size information has to be called by both approaches.
```julia
axes_map_last(x) = map(last, ArrayInterface.axes(x))
lazy_axes_map_last(x) = map(last, ArrayInterface.lazy_axes(x))

julia> @btime axes_map_last($x)
  1.608 ns (0 allocations: 0 bytes)
(4, 4, 4)

julia> @btime lazy_axes_map_last($x)
  1.470 ns (0 allocations: 0 bytes)
(4, 4, 4)
```

I've started putting this in several places and it looks like it either does as well as the eager method or improves on it.
I could probably be a little more exhaustive with improving the performance of this, but figured it was helpful enough to include as is right now.

One final note, I'm aware that I've added a bunch of messy code for bounds checking. I'm going to clean this up when I fine tune `to_indices` later on.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterface"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "3.1.9"
+version = "3.1.10"
 
 [deps]
 IfElse = "615f187c-cbe4-4ef1-ba3b-2fcf58d6d173"
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -842,10 +842,10 @@ end
 end
 
 include("ranges.jl")
-include("indexing.jl")
 include("dimensions.jl")
 include("axes.jl")
 include("size.jl")
+include("indexing.jl")
 include("stridelayout.jl")
 include("broadcast.jl")
 
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -16,8 +16,6 @@ using Base: @propagate_inbounds, tail, OneTo, LogicalIndex, Slice, ReinterpretAr
 ## utilites for internal use only ##
 _int_or_static_int(::Nothing) = Int
 _int_or_static_int(x::Int) = StaticInt{x}
-_int(i::Integer) = Int(i)
-_int(i::StaticInt) = i
 
 @static if VERSION >= v"1.7.0-DEV.421"
     using Base: @aggressive_constprop
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -37,11 +37,9 @@ function axes_types(::Type{T}) where {T}
 end
 axes_types(::Type{LinearIndices{N,R}}) where {N,R} = R
 axes_types(::Type{CartesianIndices{N,R}}) where {N,R} = R
-function axes_types(::Type{T}) where {T<:VecAdjTrans}
-    return Tuple{OptionallyStaticUnitRange{One,One},axes_types(parent_type(T), One())}
-end
-function axes_types(::Type{T}) where {T<:MatAdjTrans}
-    return eachop_tuple(_get_tuple, to_parent_dims(T), axes_types(parent_type(T)))
+function axes_types(::Type{T}) where {T<:Union{Adjoint,Transpose}}
+    P = parent_type(T)
+    return Tuple{axes_types(P, static(2)), axes_types(P, static(1))}
 end
 function axes_types(::Type{T}) where {T<:PermutedDimsArray}
     return eachop_tuple(_get_tuple, to_parent_dims(T), axes_types(parent_type(T)))
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -160,3 +160,131 @@ axes(A::Base.ReshapedArray) = Base.axes(A)  # TODO implement ArrayInterface vers
 axes(A::CartesianIndices) = A.indices
 axes(A::LinearIndices) = A.indices
 
+"""
+    LazyAxis{N}(parent::AbstractArray)
+
+A lazy representation of `axes(parent, N)`.
+"""
+struct LazyAxis{N,P} <: AbstractUnitRange{Int}
+    parent::P
+
+    LazyAxis{N}(parent::P) where {N,P} = new{N::Int,P}(parent)
+    @inline function LazyAxis{:}(parent::P) where {P}
+        if ndims(P) === 1
+            return new{1,P}(parent)
+        else
+            return new{:,P}(parent)
+        end
+    end
+end
+
+@inline Base.parent(x::LazyAxis{N,P}) where {N,P} = axes(getfield(x, :parent), static(N))
+@inline function Base.parent(x::LazyAxis{:,P}) where {P}
+    return eachindex(IndexLinear(), getfield(x, :parent))
+end
+
+@inline parent_type(::Type{LazyAxis{N,P}}) where {N,P} = axes_types(P, static(N))
+# TODO this approach to parent_type(::Type{LazyAxis{:}}) is a bit hacky. Something like
+# LabelledArrays has a linear set of symbolic keys, which could be propagated through
+# `to_indices` for key based indexing. However, there currently isn't a good way of handling
+# that when the linear indices aren't linearly accessible through a child array (e.g, adjoint)
+# For now we just make sure the linear elements are accurate.
+parent_type(::Type{LazyAxis{:,P}}) where {P<:Array} = OneTo{Int}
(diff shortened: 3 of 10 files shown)


Tokazama:
Just saw the coverage diff. I'll try to fix that up

chriselrod:
As an unrelated comment, I noticed this recently:
```julia
julia> using ArrayInterface

julia> ci = CartesianIndices(ntuple(i -> Base.OneTo(32), 4));

julia> cis = CartesianIndices(ntuple(i -> ArrayInterface.One():32, 4));

julia> @btime sum($ci)
  827.116 μs (0 allocations: 0 bytes)
CartesianIndex(17301504, 17301504, 17301504, 17301504)

julia> @btime sum($cis)
  915.828 ms (8655868 allocations: 376.07 MiB)
CartesianIndex(17301504, 17301504, 17301504, 17301504)
```
This happens when iterating over `CartesianIndices` with `OptionallyStaticUnitRange{StaticInt{F}}`. For now, I've switched `StrideArraysCore` to use `Base.OneTo` for 1-indexed axes.
The issue is a type instability caused by a type instability in [__inc](https://github.com/JuliaLang/julia/blob/d22c3041a72e62ddc3b9e17e83a61b841a6d413b/base/multidimensional.jl#L406-L416):
```julia
    @inline function __inc(state::Tuple{Int,Int,Vararg{Int,N}}, indices::Tuple{OrdinalRangeInt,OrdinalRangeInt,Vararg{OrdinalRangeInt,N}}) where {N}
        rng = indices[1]
        I = state[1] + step(rng)
        if __is_valid_range(I, rng) && state[1] != last(rng)
            return true, (I, tail(state)...)
        end
        t1, t2 = tail(state), tail(indices)
        # avoid dynamic dispatch by telling the compiler relational invariants
        valid, I = isa(t1, Tuple{Int}) ? __inc(t1, t2::Tuple{OrdinalRangeInt}) : __inc(t1, t2::Tuple{OrdinalRangeInt,OrdinalRangeInt,Vararg{OrdinalRangeInt}})
        return valid, (first(rng), I...)
    end
```
`state[1] + step(rng)` and `first(rng)` won't have the same type.
Could be fixed by making `first(::OptionallyStaticUnitRange)` always return `Int`.

Tokazama:
I'll fix that.

chriselrod:
> I approached this from a lot of different angles (some of which will probably come up in another PR still), but the issue appears to be due to the compiler not knowing to elide reconstruction of something like `Tuple{Int,Int}`, despite not undergoing any change throughout to_indices. So I'll start with the fix I had for that.

Do you have an example?
All of this should inline, in which case it should be able to decompose all the `Tuple`s into individual scalar operations.

> I chose the name because I saw a similar concept in [LoopVectorization](https://github.com/JuliaSIMD/LoopVectorization.jl/blob/2fdb50f8685c85f95e3292b137c0a8f3a9ca6c48/src/modeling/graphs.jl#L873)

It should largely be unnecessary, because the `known_*` functions should cover all the cases when it comes time for LoopVectorization to [process the loops](https://github.com/JuliaSIMD/LoopVectorization.jl/blob/3f605b94acb9c700f17166b8867316661e02082d/src/reconstruct_loopset.jl#L35-L54).
The fact that there are different methods there is somewhat a hold over from before it began using the `ArrayInterface.known_*` methods.

> A lot of array traits can be derived from looking an array's axes.
Many arrays don't have fully constructed axes stored with them, meaning we usually spend some unnecessary time composing an axis to just find out its size or offset.

This seems like it really ought to be compiled away.
Seems like a benchmarking artifact:
```julia
julia> = rand(4,5);

julia> @inline sum_first_axes(A) = sum(map(first, ArrayInterface.axes(A)))
sum_first_axes (generic function with 1 method)

julia> @btime sum_first_axes($A)
  1.410 ns (0 allocations: 0 bytes)
2

julia> @code_typed sum_first_axes(A)
CodeInfo(
1 ─     Base.arraysize(A, 1)::Int64
│       Base.arraysize(A, 2)::Int64
└──     return 2
) => Int64

julia> @code_llvm sum_first_axes(A)
;  @ REPL[23]:1 within `sum_first_axes'
define i64 @julia_sum_first_axes_1792({}* nonnull align 16 dereferenceable(40) %0) #0 {
top:
  ret i64 2
}

julia> @code_native sum_first_axes(A)
        .text
; ┌ @ REPL[23]:1 within `sum_first_axes'
        movl    $2, %eax
        retq
        nopw    %cs:(%rax,%rax)
; └
```
1 clock cycle is around 0.25ns, so anything less than that means the benchmark was completely compiled away.
Seems like that doesn't happen with the current axes in a benchmark, but that otherwise everything does disappear.

chriselrod:
I still need to spend the time to look over your other PR, but this one looks fine to me.
I'm generally in favor of being easier on the compiler. If a benchmark doesn't get compiled away with `axes` but does with `lazy_axes`, there're probably more cases where the compiler ends up doing better with the latter.

Tokazama:
> Do you have an example?
All of this should inline, in which case it should be able to decompose all the Tuples into individual scalar operations.

I always operated under that assumption until I started benchmarking everything.

```julia
# old
@btime ArrayInterface.to_indices($x, (3,3,3))
julia> @btime ArrayInterface.to_indices($x, (3,3,3))
  4.453 ns (0 allocations: 0 bytes)

# new
julia> @btime ArrayInterface.to_indices($x, (3,3,3))
  2.045 ns (0 allocations: 0 bytes)
(3, 3, 3)
```

Apparently it's a [known issue](https://github.com/JuliaLang/julia/blob/b3b9aa0eb1a90765dd3d395073992c534254a23f/base/multidimensional.jl#L799)

Tokazama:
> I still need to spend the time to look over your other PR

Just get to that when you have a chance. I recently discovered the [Data Layout Modeling](https://mlir.llvm.org/docs/DataLayout/) section of the MLIR manual and I I'm considering changing a couple of things in how I approached that.

Tokazama:
@chriselrod , is this fine to merge?

chriselrod (approved):
Yes.

PR #150, opened 2021-05-05
BigFloat/BigInt are not mutable (as arrays)

ChrisRackauckas:
The default handling works for most numbers, but not for BigFloats because they are actually `mutable struct`, but don't act like a mutable array.

Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/1400


Tokazama (commented):
LGTM

chriselrod (approved):

PR #151, opened 2021-05-11
Make axes(::Adjoint{<:Any,<:AbstractVector},1) == One():One()

chriselrod:
Before:
```julia
julia> using ArrayInterface

julia> x = rand(4);

julia> ArrayInterface.axes(x')
(Base.OneTo(1), Base.OneTo(4))
```
After:
```julia
julia> ArrayInterface.axes(x')
(1:1, Base.OneTo(4))

julia> typeof(ans)
Tuple{ArrayInterface.OptionallyStaticUnitRange{StaticInt{1}, StaticInt{1}}, Base.OneTo{Int64}}
```


Tokazama (approved):
Thanks! Can't believe I missed this first time around.

PR #153, opened 2021-05-17
Add `is_lazy_conjugate`

MasonProtter:
This adds a function `is_lazy_conjugate` that can tell if a given array is supposed to lazyily take complex conjugates or not. It's meant to understand array wrappers, but if two nested wrappers have the exact same type in a row it will get confused. Not sure if there's a way around that other than whack-a-mole.

`Adjoint` is the only lazy-`conj` wrapper I know of, but the idea is that if anyone else has such types they can register here.

### DIFF
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -841,6 +841,51 @@ end
     return setindex!(A, val; kwargs...)
 end
 
+
+"""
+    is_lazy_conjugate(::AbstractArray)  
+
+Determine if a given array will lazyily take complex conjugates, such as with `Adjoint`. This will work with
+nested wrappers, so long as there is no type in the chain of wrappers such that `parent_type(T) == T`
+
+Examples
+
+    julia> a = transpose([1 + im, 1-im]')
+    2×1 transpose(adjoint(::Vector{Complex{Int64}})) with eltype Complex{Int64}:
+     1 - 1im
+     1 + 1im
+
+    julia> ArrayInterface.is_lazy_conjugate(a)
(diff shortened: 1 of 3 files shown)


chriselrod (commented):
Some methods return `Static.True()` or `Static.False()`, which would have the advantage of letting you dispatch on whether it's true or false, and wouldn't require const prop to succeed at the user's side.

If you want to switch to `Static.True`/`Static.False`, it'd probably be worth using `@inferred` in all the tests, to make sure it succeeds.

chriselrod:
The change to `StaticBool`s should be pretty trivial:
```julia
julia> using Static

julia> !True()
False()

julia> !False()
True()
```

src/ArrayInterface.jl
```diff
@@ -841,6 +841,51 @@ end
...
+
+    julia> b = a'
+    1×2 adjoint(transpose(adjoint(::Vector{Complex{Int64}}))) with eltype Complex{Int64}:
+     1+1im  1-1im
+
+    julia> ArrayInterface.is_lazy_conjugate(b)
+    false
+
+"""
+is_lazy_conjugate(::T) where {T <: AbstractArray} = _is_lazy_conjugate(T, false)
```
chriselrod, line 869: The change to `StaticBool`s should be pretty trivial:
```julia
julia> using Static

julia> !True()
False()

julia> !False()
True()
```
That is, I think replace `false` here with `False()` is all it'd take.

MasonProtter:
Okay, I've switch to `True/False`

PR #155, opened 2021-05-23
Mark TrackedReal as immutable

ChrisRackauckas:
Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/1414 by making TrackedReal act as an immutable (like all other numbers), overriding the fact that it is a `mutable struct` but not necessarily mutable like an array.


chriselrod:
> by making TrackedReal act as an immutable (like all other numbers)

Maybe we should add a `<: Number` method?

ChrisRackauckas:
That might make sense.

PR #156, opened 2021-05-30
fix `contiguous_batch_size` for reshaped views

ranocha:
This PR fixes the following bug in ArrayInterface `v3.1.15`:
```julia
julia> using ArrayInterface

julia> u_base = randn(10, 10);

julia> u_view = view(u_base, 3, :);

julia> ArrayInterface.contiguous_batch_size(u_view) # fine
static(-1)

julia> u_reshaped_view = reshape(u_view, 1, size(u_base, 2))
1×10 reshape(view(::Matrix{Float64}, 3, :), 1, 10) with eltype Float64:
 0.8578  1.29323  0.962883  0.365737  0.480809  0.424992  -0.578577  0.28492  -0.805345  0.4239

julia> ArrayInterface.contiguous_batch_size(u_reshaped_view)
ERROR: BoundsError: attempt to access Core.SimpleVector at index [-1]
Stacktrace:
 [1] getindex
   @ ./essentials.jl:589 [inlined]
 [2] _contiguous_batch_size(#unused#::Static.StaticInt{-1}, #unused#::Tuple{Static.StaticInt{1}, Static.StaticInt{2}})
   @ ArrayInterface ~/.julia/dev/ArrayInterface/src/stridelayout.jl:266
 [3] contiguous_batch_size
   @ ~/.julia/dev/ArrayInterface/src/stridelayout.jl:263 [inlined]
 [4] contiguous_batch_size(x::Base.ReshapedArray{Float64, 2, SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}, Tuple{}})
   @ ArrayInterface ~/.julia/dev/ArrayInterface/src/stridelayout.jl:262
 [5] top-level scope
   @ REPL[6]:1
```

The implemented behavior follows the docstring
> If `contiguous_axis(T) == -1`, it will return `StaticInt{-1}()`.

This bug occurred in one of my applications of `@turbo` from LoopVectorization.jl (CC @chriselrod).

### DIFF
--- a/src/stridelayout.jl
+++ b/src/stridelayout.jl
@@ -269,6 +269,7 @@ function _contiguous_batch_size(::StaticInt{D}, ::R) where {D,R<:Tuple}
         return nothing
     end
 end
+_contiguous_batch_size(::StaticInt{-1}, ::R) where {R<:Tuple} = -One()
 
 contiguous_batch_size(::Type{Array{T,N}}) where {T,N} = Zero()
 contiguous_batch_size(::Type{BitArray{N}}) where {N} = Zero()
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -398,6 +398,12 @@ using OffsetArrays
     @test @inferred(contiguous_batch_size(@view(PermutedDimsArray(A,(3,1,2))[2:3,2,:]))) === ArrayInterface.StaticInt(-1)
     @test @inferred(contiguous_batch_size(@view(PermutedDimsArray(A,(3,1,2))[2:3,2,:])')) === ArrayInterface.StaticInt(-1)
     @test @inferred(contiguous_batch_size(@view(PermutedDimsArray(A,(3,1,2))[:,1:2,1])')) === ArrayInterface.StaticInt(0)
+    let u_base = randn(10, 10)
+        u_view = view(u_base, 3, :)
+        u_reshaped_view = reshape(u_view, 1, size(u_base, 2))
+        @test @inferred(contiguous_batch_size(u_view)) === ArrayInterface.StaticInt(-1)
+        @test @inferred(contiguous_batch_size(u_reshaped_view)) === ArrayInterface.StaticInt(-1)
+    end
 
     @test @inferred(stride_rank(@SArray(zeros(2,2,2)))) == (1, 2, 3)
     @test @inferred(stride_rank(A)) == (1,2,3)


chriselrod (approved):
Thanks, LGTM!

ranocha:
@chriselrod Using this fix, I get another error from VectorizationBase: `ERROR: MethodError: no method matching asvalbool(::Nothing)`. The basic reason for this is that `ArrayInterface.dense_dims` returns `nothing` in my application. Here is an MWE.
```julia
julia> using ArrayInterface

julia> u_base = randn(10, 10); u_view = view(u_base, 3, :); u_reshaped_view = reshape(u_view, 1, size(u_base, 2));

julia> ArrayInterface.dense_dims(u_reshaped_view)
```
The docstring of `dense_dims` is not clear enough for me. Do I understand it correctly that it should be expected that `ArrayInterface.dense_dims` returns an `NTuple{N}` for an array with `N` dimensions? If so, this behavior of `ArrayInterface.dense_dims(u_reshaped_view)` is a bug, isn't it?

chriselrod:
> So my basic question is: Is ERROR: MethodError: no method matching asvalbool(::Nothing) a bug of LoopVectorization/VectorizationBase or ArrayInterface?

I've always fixed these before by implementing the appropriate `dense_dims` method, but `val_dense_dims` should probably get a fall back that returns `false`.

`dense_dims` indicates whether that dimension is "dense".
You can think of it like this:
If a dimension is dense, you could take a slice of the array with `:` for that and all lower ranked dimensions, and `firstindex(A,i)` for all higher ranked dimensions, `vec` that slice, and then index the vector linearly / you'd have 1 solid block of memory.

So if you have:
```julia
julia> using ArrayInterface

julia> A = rand(3,4,5);

julia> vA = view(A, :, 2:4, :);

julia> ArrayInterface.dense_dims(vA)
(Static.True(), Static.True(), Static.False())
```
The first two dimensions are dense:
```julia
julia> vinds = view(LinearIndices(A), :, 2:4, :);

julia> vec(vinds[:,:,1])
9-element Vector{Int64}:
  4
  5
  6
  7
  8
  9
 10
 11
 12

julia> sort(ans) == minimum(ans):maximum(ans)
true
```

chriselrod:
The correct answer for `u_reshaped_view` would be `(False(),False()`.

You're welcome to take a stab at an implementation.
Here is another bug:
```julia
julia> u_base = randn(10, 10); u_view = view(u_base, 2:3, :); u_reshaped_view = reshape(u_view, 1, length(u_view));

julia> ArrayInterface.device(u_reshaped_view)
ArrayInterface.CPUPointer()
```
This `u_reshaped_view` is not strided, so it should not be a `CPUPointer()`, but a `CPUIndex()` (which currently is not supported by LoopVectorization).

ranocha:
Thanks for the detailed explanation, @chriselrod! I might try to dig deeper into these bugs/methods later (in another PR). I think this PR should be in a good shape and concentrates on a single thing, so it should be easier to review and merge.

chriselrod:
> I think this PR should be in a good shape and concentrates on a single thing, so it should be easier to review and merge.

Okay. Test failures are unrelated.

PR #159, opened 2021-06-01
ArrayIndex Interface

Tokazama:
The git commit provides most of the explanation for this. In the context of #141, this will provide the building blocks for layouts where there are separate index and buffer components. I'm starting with one of the simplest components so that we start off with something useful for vectorization before getting into the weeds with some other issues (change buffer types for reinterpreted arrays and adjoint arrays, when to initialize a pointer, etc.).

For review here, my main concern is if there any other components that should be included in `StrideIndex`.


chriselrod (commented):
A general comment/question I have is about the vision of the future of this library.

I'm planning/have started splitting up the JuliaSIMD libraries.
LoopVectorization currently takes a long time to precompile. Then RecursiveFactorization depends on it, DiffEqBase depend on that...
Much of the precompilation process is serial.

So I think it'd be better to split up these libraries into small pieces that strive to do only things people will likely want to use together.
For example, the pieces for adding LoopVectorization support to an array type could be in a separate library, making it a fairly light dependency for someone to take on.

Similar goes for functionality like `restructure`. If they're not generally used together, maybe we should try to put them into separate repos. Although the DiffEq ecosystem pulls everything in.

I'm planning on splitting VectorizationBase into at least four libraries:
1. Hardware summaries
2. llvmcall
3. Strided pointers
4. VectorizationBase.jl, depending on the previous three to provide the functionality requiring the others.

Maybe "3." could make use of something like this, but for as long as LoopVectorization is still using llvmcall to generate code, I may have particular requirements over how the indexing code is generated to coerce llvm into producing the assembly I want (at least for strided arrays).

Efficient code generation actually requires you don't index like `A[i,j,k]`, but something closer to:
```julia
# U1, U2, and U3 are unroll factors 1, 2, and 3
ptra = pointer(A)
for k in K
    for j in J
        for i in I
            # incrementing the strided pointer instead of the raw pointer results in worse codegen
            # hence, increment the raw pointer and only assemble the strided pointer as it is used
            # to save on verbosity, pretend this load and store and unrolled by U1, U2, and U3
            # with appropriate offset calculations
            vload(assemble_strided_pointer(ptra, strideinfo...))
            vstore!(assemble_strided_pointer(ptra, strideinfo...), v)
            ptra += stride(A,1)*sizeof(eltype(A))*U1
        end
        ptra += stride(A,2)*sizeof(eltype(A))*U2
    end
    ptra += stride(A,3)*sizeof(eltype(A))*U3
end
```
`StrideIndex` can be used for this, if you think about it in terms of calculating offsets instead of indexing.

On the other hand, that risks fragmenting the ecosystem and hindering discoverability for the sake of working around Julia being slow to (pre)compile.

Something else that'd be nice to have for alternate data structures is a generalization of `ArrayInterface.contiguous_axis`. That is, some arrays may have contiguous elements you can efficiently load/store, but that may not be aligned with an axis. Like the `(/Bi/Tri)Diagonal`.

Tokazama:
> So I think it'd be better to split up these libraries into small pieces that strive to do only things people will likely want to use together.

It seems natural to me to have the generic traits, indexing, and array reconstruction stuff here; but ideally this is all scaffolding for other packages to build on. I'd like to get support for the basic arrays we've be working on in "stridelayout.jl" for this, but after that I think it would make sense for there to be other packages hosting some of this code (like the block index types, but I didn't implement those and don't know who would be interested in packaging them). I'm also a bit hesitant to spend much energy supporting all of the arrays found in `LinearAlgebra` and `SparseArrays`. I'm hopeful I can also get rid of a good chunk of private code here once the layouts stuff is in place.

> StrideIndex can be used for this, if you think about it in terms of calculating offsets instead of indexing.

That's probably a more accurate description of what I'm going for here.
I've been working on something that works a bit like this:

```julia
function index_plan(::Type{T}, ::IndexLinear)
    # p is callable and produces an ArrayIndex for parent_type(T) given the indexing style it is provided
    p = index_plan(
        parent_type(T),
        IndexStyle(T)  # the type of index passed to the parent
    )
    combine_index_plans(T, p) # attempt to combine p with information about T
end
```

With the final output called on an instance of `T`.
This allows us to change the layout depending on the `IndexStyle` (e.g., `UnorderedIndex` for summation). 
Composing the raw buffer and pointer would be separate entirely, which I'm assuming would be in line with separate assembly of a strided pointer.
I think it makes more sense for people to just use StrideArrays.jl than for use to implement a strided pointer here, but I don't see anyway around managing pointers when dealing with reinterpreted arrays.

Tokazama:
I took some time to clean up some of the code I've been working on so that the direction I'm trying to go with this is a little bit more clear. [This gist](https://gist.github.com/Tokazama/8e0017a0084cfcbfaa556d754aebfa5a.) has most of the components that aren't implemented here yet (and maybe a few superfluous things I'll get rid of before submitting anything). Building on the example you used, a somewhat naive version with layouts might look like this:

```julia
fxn(a::A) where {A} = fxn(a, materialize_layout(layout(A, CartesianAccess()), a))
function fxn(a::A, lyt::ArraIndex) where {A}
    ptra = pointer(a)
    for k in K
        for j in J
            for i in I
                index = lyt[k, j, i]
                unsafe_load(ptra, index)
                unsafe_store!(ptra, v, index)
            end
        end
    end
end
```

Admittedly, this wouldn't do nearly as well as your example.
We could have `unsafe_load` and `unsafe_store!` dispatch on a unique pointer type derived from something like StrideArray, but I doubt that will close the gap here.
The goal with layouts is to provide multiple opportunities for contextual optimizations when accessing array data.
Because `CartesianAccess` (subtype of `AccessStyle`) is propagated to the mostly deeply nested array, each layer of `A` has the opportunity to produce some preliminary method call.
Once the most deeply nested array is reached each layer attempts to combine methods with its parent until reaching the outermost layer.
If the core buffer for an array provides unique information about hardware then this can be propagated to the final methods call.
Propagating this information would involve defining a new `combine_layouts` method and we could probably use a more informative `AccessStyle` too.
We could the have the following:

```julia
function combine_layouts(::Type{X}, ::Type{Y}) where {X,Y<:UnrolledIndex}
    ...
end
fxn(a::A) where {A} = fxn(a, materialize_layout(layout(A, LoopAccess{3}()), a))
function fxn(a::A, lyt::UnrolledIndex) where {A}
    ptra = pointer(a)
    for k in K
        for j in J
            for i in I
                vload(assemble_strided_pointer(ptra, strideinfo...))
                vstore!(assemble_strided_pointer(ptra, strideinfo...), v)
                ptra += unroll(lyt, 1)   # == stride(A,1)*sizeof(eltype(A))*U1
            end
            ptra += unroll(lyt, 2)       # == stride(A,2)*sizeof(eltype(A))*U2
        end
        ptra += unroll(lyt, 3)           # == stride(A,3)*sizeof(eltype(A))*U3
    end
end
```

I figured that `StrideIndex` and then a bare bones layout interface would be a good starting point, instead of bikeshedding a bunch of code and finding out that none if it is particularly useful.

chriselrod:
As an aside, before splitting up VectorizationBase, I'd like to just make Polyster.jl no longer depend on VectorzationBase.jl.
I've already done this for ThreadingUtilities, but step 2 is StrideArraysCore.
It defines a `PtrArray` as a `VectorizationBase.StridedPointer` + `ArrayInterface.size` tuple.

So that would be a good use case for the indexing and layout infrastructure here, i.e. drop `VectorizationBase` and use the code you've been adding here.

> I'm also a bit hesitant to spend much energy supporting all of the arrays found in `LinearAlgebra` and `SparseArrays`. I'm hopeful I can also get rid of a good chunk of private code here once the layouts stuff is in place.

One of my concerns there is that I think we should have a use case to motivate the implementation before creating the implementation.
When I start adding things like `Symmetric`/`Hermitian` support to LoopVectorization, I'll be able to better reason about what I want related code supporting a `HermitianStridedPointer` or a layout description to look like.
I'm still thinking it'd be a good idea to switch to LLVM.jl as a backend instead of creating a giant Julia `Expr`, so LoopVectorization itself may no longer need `StridedPointer`s itself.
Unfortunately, the `Symmetric` wrapper isn't typed based on whether it is `L` or `U`, but more generally, a front end integrated with a compiler should be able to figure some things out like `A[i,j] == parent(A)[j,i]` if `i > j`, for example. I'll have to get that far before realizing what I do want help with.

> I would love to have a small and extensible package (similar to Static.jl) that provides methods/types for allocating memory.

My number one want regarding memory allocation is adding an `alloca` intrinsic to Julia base.
[Here](https://julialang.zulipchat.com/#narrow/stream/137791-general/topic/StrideArrays/near/237898198) is a Zulip discussion, in case you missed it.
I should probably post to the internals slack and ask what we need.
Aside from being dynamically sized, this is more ergonomic IMO than `mutable struct MemoryBuffer{N,T}; data::NTuple{N,T}; end` + `GC.@preserve`.

This would give us dynamically sized stack allocated arrays (where the memory is freed when the function that called `alloca` returns).

For statically sized arrays, `MemoryBuffer` is fine, and has the benefit of letting arrays actually be returned if you want (at the cost of heap allocating the memory).
Otherwise, I'm taking the approach of making a large `Vector` and writing my code to use this.
[UnPack.jl](https://github.com/mauro3/UnPack.jl) makes this much more convenient. You can define `UnPack.unpack` and `UnPack.pack!` methods to construct your `PtrArray`s and/or load/store your fields for you, to hide all the ugliness and make the resulting code still look clean/Julian.


> Also, feel free to ping me if you just need someone to look over a PR in JuliaSIMD. I'm pretty busy with an MRI study right now, but I usually manage to check on github throughout the day and might be able to help every once in a while.

Sure. I'll ping you when I make the StrideArraysCore PR dropping VectorizationBase. Although you're welcome to make that PR yourself if you want to make a proof of concept of ArrayInterface's layout representations. You wrote that code and are of course much more familiar with it than I am.

> [This gist](https://gist.github.com/Tokazama/8e0017a0084cfcbfaa556d754aebfa5a.) has most of the components that aren't implemented here yet (and maybe a few superfluous things I'll get rid of before submitting anything).

I can't view this gist.

> We could have `unsafe_load` and `unsafe_store!` dispatch on a unique pointer type derived from something like StrideArray, but I doubt that will close the gap here.

> a somewhat naive version with layouts might look like this:

Something like this would be somewhat similar to what LoopVectorization currently wants:
```julia
fxn(a::A) where {A} = fxn(a, materialize_layout(layout(A, CartesianAccess()), a))
function fxn(a::A, lyt::ArraIndex) where {A}
    ptra = pointer(a)
    offsetk, offsetj, offseti = offsets(lyt)
    for k in K
        for j in J
            for i in I
                unsafe_load(ptra)
                unsafe_store!(ptra, v)
                ptra += lyt[offsetk, offsetj, i]
            end
            ptra += lyt[offsetk, j, offseti]
        end
        ptra += lyt[k, offsetj, offseti]
    end
end
```

On dropping VectorizationBase, I realized recently that for the actual `getindex`/`setindex!` methods, we should avoid `llvmcall`: https://github.com/JuliaSIMD/VectorizationBase.jl/issues/55
Meaning `unsafe_load`/`unsafe_store!` is the way to go. Although probably with a convenience wrapper:
https://github.com/JuliaSIMD/ThreadingUtilities.jl/blob/d76942c42bb98820428027ff721404f4d986a485/src/utils.jl#L1-L16
Maybe we should move this somewhere, such as to the small memory library you mentioned.
I find lots of places wanting it that don't necessarily want to depend on each other otherwise.

Regarding unrolled index, there is
```julia
julia> using VectorizationBase

help?> VectorizationBase.Unroll
  Unroll{AU,F,N,AV,W,M,X}(i::I)

    •  AU: Unrolled axis

    •  F: Factor, step size per unroll. If AU == AV, F == W means successive loads. 1 would mean offset by 1, e.g. x{1:8], x[2:9], and x[3:10].

    •  N: How many times is it unrolled

    •  AV: Vectorized axis # 0 means not vectorized, some sort of reduction

    •  W: vector width

    •  M: bitmask indicating whether each factor is masked

    •  X: stride between loads of vectors along axis AV.

    •  i::I - index
```
(The doc string shouldn't have extra spaces between the bullets; I should figure out how to get rid of them...)

This is used by `vload`/`vstore!` to describe unrolled loads/stores, because a lot of optimizations are possible for certain kinds of unrolled loads/stores (i.e., if the vectors aren't contiguous in memory, but the contiguous axis is unrolled, you can use shuffles to make the load/store much more efficient).
Maybe that should be made to support incrementing `ptr`s as well, but LoopVectorization's more manual approach works for now, and longer term I want to replace the backend/code generation all together to try and really cut down latency.

chriselrod:
https://github.com/JuliaSIMD/ManualMemory.jl

Tokazama:
Fixed [gist link](https://gist.github.com/Tokazama/8e0017a0084cfcbfaa556d754aebfa5a)

chriselrod (approved):
I'll go ahead and approve this rather than leave it in limbo forever.

chriselrod:
Then we can make a new ArrayInterface release, and I'll redefine `PtrArray` to be a `Ptr`, `StrideIndex`, and `size` tuple.

Issue #160, opened 2021-06-02
Bug in `strides` for reshaped views


ranocha:
```julia
julia> using ArrayInterface

julia> u_base = randn(10, 10); u_view = view(u_base, 3, :); u_reshaped_view1 = reshape(u_view, 1, :); u_reshaped_view2 = reshape(u_view, 2, :);

julia> ArrayInterface.strides(u_reshaped_view1)
(static(1), 1)

julia> ArrayInterface.strides(u_reshaped_view2)
(static(1), 2)
```
These seem to be wrong, I think. Because of `ArrayInterface.strides(u_view) == (10,)`, the `strides` should be
- `ArrayInterface.strides(u_reshaped_view1) = (10 #= doesn't matter since there is only one element in this dimension =#, 10)`
- `ArrayInterface.strides(u_reshaped_view1) = (10, 20)`
Is that correct?

ranocha:
I will give it a shot.

Tokazama:
I don't think we can actually guarantee that a `ReshapedArray` can combine with its parent strides unless we know that the the reshaped dimensions are contiguous and dense, because otherwise we would need to split strides. For example:

```julia
x = Array{Int}(undef, 5, 2, 5, 2); 
xv = view(x, :, 1, :, :);
xvr = reshape(xv, 10, 5)
```

If reshaping was done with static information we could figure this out at compile time, but with the current design of `ReshapedArray` we would have to compute all the parent strides and then see if one of the new dimensions is large enough to split parent strides.

ranocha:
Yeah, that's like https://github.com/JuliaArrays/ArrayInterface.jl/issues/162#issuecomment-853109218

Issue #157, opened 2021-05-31
Bug in `dense_dims` for reshaped views


ranocha:
Reported in https://github.com/JuliaArrays/ArrayInterface.jl/pull/156#issuecomment-851002133:
```julia
julia> using ArrayInterface

julia> u_base = randn(10, 10); u_view = view(u_base, 3, :); u_reshaped_view = reshape(u_view, 1, size(u_base, 2));

julia> ArrayInterface.dense_dims(u_reshaped_view)
```
As confirmed by @chriselrod in https://github.com/JuliaArrays/ArrayInterface.jl/pull/156#issuecomment-851293134, this is a bug:
> The correct answer for `u_reshaped_view` would be `(False(),False())`.

Tokazama:
resolved with #161

ranocha:
I fixed it only for the case of reshaped versions of abstract vectors. It still fails for reshaped views of higher-dimensional arrays, e.g.,
```julia
(@v1.7) pkg> activate --temp
  Activating new project at `/tmp/jl_4vH1Gz`

(jl_4vH1Gz) pkg> add ArrayInterface
    Updating registry at `~/.julia/registries/General.toml`
   Resolving package versions...
    Updating `/tmp/jl_4vH1Gz/Project.toml`
  [4fba245c] + ArrayInterface v6.0.14

julia> using ArrayInterface

julia> u_base = randn(10, 10, 10); u_view = view(u_base, 3, :, :); u_reshaped_view = reshape(u_view, 1, size(u_base, 2) * size(u_base, 3));

julia> ArrayInterface.dense_dims(u_reshaped_view)

julia> ArrayInterface.dense_dims(u_reshaped_view) |> isnothing
true
```

Tokazama:
I guess in that example we could know that no dimensions are dense because `u_view` doesn't have any dense dimensions, but I'm not sure we can do any better at identifying instances where they are dense.

Tokazama:
#307 tests all the examples in this issue. I'm assuming there are still times when we don't detect dense dimensions just because we can't know in some situations without knowing the size.

PR #161, opened 2021-06-02
reshaped views and reinterpreted arrays

ranocha:
I have fixed the issues reported in #157 and #160. Here, I concentrated on the exact applications using reshaped versions of views that are abstract vectors. In this case, perfect might be the enemy of good again: This solution isn't perfect but it solves the problems I have right now and can be extended later to reshaped arrays of more general array types. I also adapted some docstrings slightly to look more accurate to me.

### DIFF
--- a/src/stridelayout.jl
+++ b/src/stridelayout.jl
@@ -88,7 +88,7 @@ end
     contiguous_axis(::Type{T}) -> StaticInt{N}
 
 Returns the axis of an array of type `T` containing contiguous data.
-If no axis is contiguous, it returns `StaticInt{-1}`.
+If no axis is contiguous, it returns a `StaticInt{-1}`.
 If unknown, it returns `nothing`.
 """
 contiguous_axis(x) = contiguous_axis(typeof(x))
@@ -297,7 +297,7 @@ contiguous_batch_size(::Type{<:Base.ReinterpretArray{T,N,S,A}}) where {T,N,S,A} 
 """
     is_column_major(A) -> True/False
 
-Returns `Val{true}` if elements of `A` are stored in column major order. Otherwise returns `Val{false}`.
+Returns `True()` if elements of `A` are stored in column major order. Otherwise returns `False()`.
 """
 is_column_major(A) = is_column_major(stride_rank(A), contiguous_batch_size(A))
 is_column_major(sr::Nothing, cbs) = False()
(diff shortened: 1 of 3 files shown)


src/stridelayout.jl
```diff
@@ -310,10 +310,11 @@ _is_column_major(sr::R, cbs::StaticInt) where {R} = False()
 _is_column_major(sr::R, cbs::Union{StaticInt{0},StaticInt{-1}}) where {R} = is_increasing(sr)
 
 """
-    dense_dims(::Type{T}) -> NTuple{N,Bool}
+    dense_dims(::Type{<:AbstractArray{N}}) -> NTuple{N,Bool}
```
chriselrod, line 313: `NTuple{N,StaticBool}`.

src/stridelayout.jl
```diff
@@ -485,6 +492,17 @@ function strides(x)
         return Base.strides(x)
     end
 end
+
+# Fixes the example of https://github.com/JuliaArrays/ArrayInterface.jl/issues/160
+# TODO: Should be generalized to reshaped arrays wrapping more general array types
+function strides(A::ReshapedArray{T,N,P}) where {T, N, P<:AbstractVector}
+    if defines_strides(A)
+        return size_to_strides(size(A), first(strides(parent(A))))
```
chriselrod, line 500: What about
```julia
julia> A = rand(5,12);

julia> B = reshape(A', (3,4,5));
```
`ArrayInterface.strides(B)` should be `(5,15,StaticInt{1}())`.

chriselrod (commented):
Thanks for tackling this!

Re "don't let perfect be the enemy of good", if you can't address the comment on reshaping `PermutedDimsArray`/`transpose`/`adjoint` at the moment, we can merge this now and file a new issue.

src/stridelayout.jl
ranocha: I would prefer to fix that later since I would need to think a bit more about it. I opened https://github.com/JuliaArrays/ArrayInterface.jl/issues/162 to track this bug.

src/stridelayout.jl
ranocha: Thanks! Implemented in 7335b49

ranocha:
I pushed a fix for #163. I also updated the version in Project.toml since I would really like to have a new release to be able to use these fixes.

chriselrod (approved):
LGTM.

PR #164, opened 2021-06-03
Fix ismutable for Julia 1.7

chriselrod:
I arbitrarily picked a Julia version for the comparison.


chriselrod:
Tests still fail on Julia nightly because of ambiguities in `lu(::StaticArray)` methods.

ChrisRackauckas (approved):

PR #167, opened 2021-06-05
Static axis 1 for adj or trans

chriselrod:
With this PR, it'll return `One():One()` only when called with `StaticInt` on the axis.
Should we have it return `One():One()` also for `1`?
The concern there is that we'd rely on const prop for type instability.

I've similarly extended it to return `One():One()` for `::StaticInt{M}` where `M` is greater than the array's dimension.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterface"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "3.1.16"
+version = "3.1.17"
 
 [deps]
 IfElse = "615f187c-cbe4-4ef1-ba3b-2fcf58d6d173"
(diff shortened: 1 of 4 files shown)


Tokazama:
When I was originally going through this code I thought about this and decided that if we return a statically sized axis for SArray then it makes sense to return a statically sized axis for these N+ dimensions.

However, this is probably a better solution since the current behavior might be unexpected for non static arrays and we can still get the static axis with StaticInt

Tokazama (approved):

chriselrod:
We could still consider that later, "if someone doesn't want statically sized axes, they could use `Base.axes` instead".
But I wanted a relatively minimal change for now.

`LoopVectorization.@turbo` replaces `axes(A,1)` and `indices(A,1`) with `ArrayInterface.axes(A, StaticInt(1))` , so `StaticInt` support covers my use cases.

PR #176, opened 2021-07-30
clean up range code and documentation

Tokazama:
* Cleaned up messy code so that it's easier to read and more concise
* Previously test code in doc strings weren't in code blocks. Reimplemented them as comparable examples.
* Swapped `kwargs.data` to `values(kwargs)` in accordance to deprecation warning in Julia v1.7.

### DIFF
--- a/src/ranges.jl
+++ b/src/ranges.jl
@@ -5,8 +5,13 @@
 If `first` of an instance of type `T` is known at compile time, return it.
 Otherwise, return `nothing`.
 
-@test isnothing(known_first(typeof(1:4)))
-@test isone(known_first(typeof(Base.OneTo(4))))
+```julia
+julia> ArrayInterface.known_first(typeof(1:4))
+nothing
+
+julia> ArrayInterface.known_first(typeof(Base.OneTo(4)))
+1
+```
 """
 known_first(x) = known_first(typeof(x))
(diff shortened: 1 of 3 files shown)


src/ranges.jl
```diff
@@ -65,20 +81,15 @@ at compile time. An `OptionallyStaticUnitRange` is intended to be constructed in
 from other valid indices. Therefore, users should not expect the same checks are used
 to ensure construction of a valid `OptionallyStaticUnitRange` as a `UnitRange`.
 """
-struct OptionallyStaticUnitRange{F<:Integer,L<:Integer} <: AbstractUnitRange{Int}
+struct OptionallyStaticUnitRange{F<:CanonicalInt,L<:CanonicalInt} <: AbstractUnitRange{Int}
```
chriselrod, line 84: Is this a breaking change / is it likely any packages in the ecosystem actually used anything other than a `CanonicalInt`?

chriselrod (approved):
I think this is fine. I doubt anyone was using `(U)Int128` or `BigInt` here.

src/ranges.jl
Tokazama: This was being forced through construction previously so there shouldn't be any change in funciton.

PR #177, opened 2021-08-01
Docs Workflow, README, and ArrayIndex boundschecks

Tokazama:
* Added workflow for docs
* Removed the docs from README so they could stay up to date
* Added `@boundscheck` to `MatrixIndex` subtypes


chriselrod (approved):
Thanks!

PR #183, opened 2021-08-06
more `@static if`'s on version checks

ChrisRackauckas:


ChrisRackauckas:
pretty uncontroversial.

Issue #182, opened 2021-08-03
ArrayInterface.size vs StaticArrays.Size


cscherrer:
Hi,

I'd like to get a better understanding of the intending idioms for working with arrays, especially the case where we want different methods for the static vs dynamic case.

`StaticArrays.Size` solves the problem like this:
```julia
julia> Size(zeros(SMatrix{3, 4}))
Size(3, 4)

julia> Size(zeros(3, 4))
Size(StaticArrays.Dynamic(), StaticArrays.Dynamic())
```

Though we might have some missing results, it's nice that the size computation itself is always static. Working with types is also convenient:
```julia
julia> Size(typeof(zeros(SMatrix{3, 4})))
Size(3, 4)

julia> Size(typeof(zeros(3, 4)))
Size(StaticArrays.Dynamic(), StaticArrays.Dynamic())
```
By contrast, here's `ArrayInterface.size`:
```julia
julia> ArrayInterface.size(zeros(SMatrix{3, 4}))
(static(3), static(4))

julia> ArrayInterface.size(zeros(3, 4))
(3, 4)
```

That last result might seem convenient, but computing this requires dynamic dispatch, which seems to defeat the purpose. And it fails on types:

```julia
julia> ArrayInterface.size(typeof(zeros(3, 4)))
ERROR: MethodError: no method matching size(::Type{Matrix{Float64}})
Closest candidates are:
  size(::Union{LinearAlgebra.QR, LinearAlgebra.QRCompactWY, LinearAlgebra.QRPivoted}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/qr.jl:524
  size(::Union{LinearAlgebra.QR, LinearAlgebra.QRCompactWY, LinearAlgebra.QRPivoted}, ::Integer) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/qr.jl:523
  size(::Union{LinearAlgebra.Cholesky, LinearAlgebra.CholeskyPivoted}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/cholesky.jl:442
  ...
Stacktrace:
 [1] axes
   @ ./abstractarray.jl:89 [inlined]
 [2] axes
   @ ~/.julia/packages/ArrayInterface/CmPZc/src/axes.jl:166 [inlined]
[...]

PR #184, opened 2021-08-07
Document static traits (#182)

Tokazama:


Tokazama:
@cscherrer, do you think this documentation appropriately covers what was discussed in #182?

docs/src/index.md
```diff
@@ -11,3 +11,38 @@ CurrentModule = ArrayInterface
 Modules = [ArrayInterface]
 ```
 
+## Static Traits
+
+The size along one or more dimensions of an array may be know at compile time.
```
cscherrer, line 16:
```suggestion
The size along one or more dimensions of an array may be known at compile time. 
```

docs/src/index.md
```diff
@@ -11,3 +11,38 @@ CurrentModule = ArrayInterface
...
+```julia
+fxn(x) = _fxn(ArrayInterface.size(x), x)
+_fxn(sz::Tuple{StaticInt{S1},StaticInt{S2}}, x) where {S1,S2} = ...
+_fxn(sz::Tuple{StaticInt{3},StaticInt{3}}, x) = ...
+_fxn(sz::Tuple{Int,StaticInt{S2}}, x) where {S2} = ...
+_fxn(sz::Tuple{StaticInt{S1},Int}, x) where {S1} = ...
+_fxn(sz::Tuple{Int,Int}, x) = ...
+```
+
+Generally, `ArrayInterface.size` uses the return of `known_size` to form a static value for those dimensions with known length and only queries dimensions corresponding to `nothing`.
```
cscherrer, line 41: If an array has some known dimensions and some unknown, will `ArrayInterface.size` force a dynamic dispatch?

docs/src/index.md
```diff
@@ -11,3 +11,38 @@ CurrentModule = ArrayInterface
...
+For example, the previous example had a known size of `(1, nothing)`.
+Therefore, `ArrayInterface.size` would have compile time information about the first dimension returned as `static(1)` and would only look up the size of the second dimension at run time.
+Generic support for `ArrayInterface.known_size` relies on calling `known_length` for each type returned from `axes_types`.
+Therefore, the recommended approach for supporting static sizing in newly defined array types is defining a new `axes_types` method.
```
cscherrer, line 45: This is really helpful!

cscherrer (approved):
Hi @Tokazama , thanks for pinging me on this. This is really coming along! I added some comments and found one small typo.

I really like the recommendation in line 41, and in general it would be great to have more of these. I don't know if how established best practices are for this sort of thing, but it still takes a lot of thinking to write generic code that dispatches to specialized static code when possible. So any recommendations or toy examples would be a really big help.

I've thought a little more about framing this problem; I'll add to the discussion in https://github.com/JuliaArrays/ArrayInterface.jl/issues/182

docs/src/index.md
Tokazama: The only way dispatch is dynamic is if you try to enforce staticness like this:
```julia
fxn(x) = _fxn(Static.static(ArrayInterface.size(x)), x)
_fxn(sz::Tuple{StaticInt{S1},StaticInt{S2}}, x) where {S1,S2} = ...
```

Otherwise the compiler should know that something like `_fxn(sz::Tuple{StaticInt{S1},Int}, x) where {S1} = ...` is the correct method.

Issue #187, opened 2021-08-11
Strides on when SubArray has a scalar sub index


Tokazama:
I've been getting incorrect results when using scalar indexing that doesn't start with one on views.

```julia
julia> A = zeros(3, 4, 5);

julia> A[:] = 1:60;

julia> Aview = view(A, :, 2, :);

julia> ArrayInterface.StrideIndex(Aview)[1,1]
1

julia> Aview[1,1]
4.0

julia> Aview = view(A, 2, :, :);

julia> ArrayInterface.StrideIndex(Aview)[1,1]
1

julia> Aview[1,1]
2.0

```

Should we be combining those scalar indices with strides?

chriselrod:
```julia
julia> A = zeros(3, 4, 5);

julia> A[:] = 1:60;

julia> Aview = view(A, :, 2, :);

julia> ArrayInterface.StrideIndex(Aview)[1,1]
1

julia> Aview[ArrayInterface.StrideIndex(Aview)[1,1]]
4.0

julia> Aview[1,1]
4.0
```
What's the desired behavior here?
I guess (for efficiency), that it returns a linear index into the memory of the array?

In that case, the `StrideIndex` object should carry an offset.
The offset should default to `Zero()` for most array types, for subarrays should be calculated using the offsets implied by the non-colon indices. E.g., `(2 - offset(A,2)) * stride(A,2)` for the first example, and `(2 - offset(A,1)) * stride(A,1)` for the second.

chriselrod:
Also, methods like `pointer` do the offsetting for you:
```julia
julia> using ArrayInterface

julia> A = zeros(3, 4, 5);

julia> A[:] = 1:60;

julia> Aview = view(A, :, 2, :);

julia> unsafe_load(pointer(Aview), ArrayInterface.StrideIndex(Aview)[1,1])
4.0

julia> Aview[1,1]
4.0

julia> Aview = view(A, 2, :, :);

julia> unsafe_load(pointer(Aview), ArrayInterface.StrideIndex(Aview)[1,1])
2.0

julia> Aview[1,1]
2.0
```

PR #189, opened 2021-08-17
Add Equal() to indicate that offset1 equals first offset

chriselrod:
Save a value if we cross a function boundary with `OffsetVector`s.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterface"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "3.1.23"
+version = "3.1.24"
 
 [deps]
 IfElse = "615f187c-cbe4-4ef1-ba3b-2fcf58d6d173"
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -944,6 +944,9 @@ function __init__()
         @inline axes(A::OffsetArrays.OffsetArray{T,N}, ::StaticInt{M}) where {T,M,N} = _axes(A, StaticInt{M}(), gt(StaticInt{M}(),StaticInt{N}()))
         @inline known_offset1(::Type{<:OffsetArrays.OffsetArray}) = 1
         @inline known_offset1(::Type{<:OffsetArrays.OffsetVector}) = nothing
+        function StrideIndex{N,R,C}(a::OffsetArrays.OffsetVector) where {N,R,C}
+            return StrideIndex{N,R,C}(strides(a),offsets(a),Equal())
+        end
     end
 end
 
--- a/src/array_index.jl
+++ b/src/array_index.jl
@@ -184,6 +184,19 @@ function BandedBlockBandedMatrixIndex(
 end
 
 """
+  Equal
+
+Equality indicator. Left generic to indicate possible other uses where it's easy to interpret.
+In case of `(s::StrideIndex).offset1 === Equal()`, this means that `offset1` is
+```julia
+s.offset[1]
+```
+
+Use of `StaticInt`s is preferred when possible for convenience. The purpose here is simply
+compressing runtime size information for when `StrideIndex` is passed as a non-inlined argument.
+"""
+struct Equal end
+"""
     StrideIndex(x)
 
 Subtype of `ArrayIndex` that transforms and index using stride layout information
--- a/src/stridelayout.jl
+++ b/src/stridelayout.jl
@@ -74,7 +74,8 @@ known_offset1(x) = known_offset1(typeof(x))
 known_offset1(::Type{T}) where {T} = _known_offset1(has_parent(T), T)
 _known_offset1(::True, ::Type{T}) where {T} = known_offset1(parent_type(T))
 _known_offset1(::False, ::Type{T}) where {T} = 1
-known_offset(::Type{<:StrideIndex{N,R,C,S,O,O1}}) where {N,R,C,S,O,O1} = known(O1)
+known_offset1(::Type{<:StrideIndex{N,R,C,S,O,O1}}) where {N,R,C,S,O,O1} = known(O1)
+known_offset1(::Type{<:StrideIndex{N,R,C,S,O,Equal}}) where {N,R,C,S,O} = first(known(O))
 
 """
     offset1(x) -> Union{Int,StaticInt}
@@ -90,6 +91,7 @@ Returns the offset of the linear indices for `x`.
     end
 end
 offset1(x::StrideIndex) = getfield(x, :offset1)
+offset1(x::StrideIndex{N,R,C,S,O,Equal}) where {N,R,C,S,O} = getfield(getfield(x, :offsets), 1, false)
 
 """
     contiguous_axis(::Type{T}) -> StaticInt{N}
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -697,6 +697,10 @@ end
         @test @inferred(ArrayInterface.known_offsets(ap_index)) === ArrayInterface.known_offsets(Ap)
         @test @inferred(ArrayInterface.known_offset1(ap_index)) === ArrayInterface.known_offset1(Ap)
         @test @inferred(ArrayInterface.known_strides(ap_index)) === ArrayInterface.known_strides(Ap)
+        
+        @test @inferred(ArrayInterface.known_offset1(o)) === nothing
+        @test @inferred(ArrayInterface.StrideIndex(o)).offset1 === ArrayInterface.Equal()
+        @test @inferred(ArrayInterface.offset1(o)) === @inferred(ArrayInterface.offset1(ArrayInterface.StrideIndex(o)))
     end
 
     if VERSION ≥ v"1.6.0-DEV.1581"


Tokazama:
Would it make sense to have `offset1` also account for offsets in `SubArray` (e.g., #187). Then the only case where `offset1` wouldn't be `Equal` would be when it represents a `SubArray`.

chriselrod:
I'd prefer not to have `offset1` handle the `SubArray`'s offset, because then I'd need to special case `stridedpointer` on `SubArray` to take the pointer of the parent.

I'm transitioning `StridedPointer`s to use `StrideIndex` in [LayoutPointers](https://github.com/JuliaSIMD/LayoutPointers.jl), and will then eventually have VectorizationBase depend on it.

Tokazama:
> I'd prefer not to have offset1 handle the SubArray's offset, because then I'd need to special case stridedpointer on SubArray to take the pointer of the parent.

Isn't this done with `memory_reference` [here](https://github.com/JuliaSIMD/LayoutPointers.jl/blob/7a51a5ecfa6b681a4aae1dc298bd48710bcddde1/src/stridedpointers.jl#L25)? Side note, wouldn't you have `PseudoPtr` returned with the same offset when the device is `CPUIndex`?

If `offset1(x)` doesn't refer to the linear offset to the first value of `x`, then should the default definition be `offset1(x) = getfield(offsets(x), 1, false)` and we drop the `offset1` field entirely from `StrideIndex`?

Tokazama:
Sorry if my comment seems off topic. It seems to me that this PR is improving upon a redundancy (or ambiguity) of `offset1` (the method and field of `StrideIndex`). I just want to be sure that we formalize what `offset1` is before creating additional layers of complexity on top of it.

chriselrod:
> Isn't this done with `memory_reference` [here](https://github.com/JuliaSIMD/LayoutPointers.jl/blob/7a51a5ecfa6b681a4aae1dc298bd48710bcddde1/src/stridedpointers.jl#L25)? Side note, wouldn't you have `PseudoPtr` returned with the same offset when the device is `CPUIndex`?

Ha, forgot I was already special casing it.
So making that change would involve removing this special casing (so that `memory_refernce` looks to the parent), and then make sure I set its offsets correctly.

For me, it is easier not to make those changes. LV tries to zero the offsets, meaning incrementing the pointers, anyway, so w/ respect to LV, both versions should compile to the same code anway.
I think `A[StrideIndex(A)[inds...]] == A[inds...]` also makes sense, which we have from the current definition, while the new one would have `parent(A)[StrideIndex(A)[inds...]] = A[inds...]`.

> If `offset1(x)` doesn't refer to the linear offset to the first value of `x`, then should the default definition be `offset1(x) = getfield(offsets(x), 1, false)` and we drop the `offset1` field entirely from `StrideIndex`?

IMO, `unsafe_load(pointer(A)) == first(A)`, which is true for `SubArray`s because `pointer` (and `memory_reference`) offset the pointer when called so that this invariant holds.

Under the perspective that the base pointer of a `SubArray` is the pointer to its first element (rather than the pointer to the parent), then `offset1` currently does return the linear offset.
And for `A::OffsetArray{T,N}` where `N > 1`, we do not have that `offset1(A) == first(offsets(A))`.

chriselrod:
> I've been hacking on a set of `ArrayIndex` subtypes for `ReshapedArray`, `PermutedDimsArray`, `SubArray`, and `ReinterpretArray`. Then these can be arbitrarily combined (with something like `combined_index(child_index, parent_index)`). The first three are pretty straightforward but I think we'd need to have some way of representing N bytes per each index (which shouldn't be that hard, but I haven't had time to think through a proper solution all the way).

If these arrays are all strided, I'd rather have a single representation capable of describing any of them. `StrideIndex` with rank `R` should be able to handle any of these (assuming they're what ArrayInterface considers to be `StridedArray`s).

chriselrod:
> Sorry if my comment seems off topic. It seems to me that this PR is improving upon a redundancy (or ambiguity) of `offset1` (the method and field of `StrideIndex`). I just want to be sure that we formalize what `offset1` is before creating additional layers of complexity on top of it.

I am happy with the current semantics, I just wanted to elliminate redundant data storage.
LoopVectorization tries to pass data in registers instead of on the stack, but the number of registers you can use for this is very limited.

Tokazama:
Neither `A[StrideIndex(A)[inds...]]` nor `parent(A)[StrideIndex(A)[inds...]` is technically the current way this functions.

For example, something we currently test with `StrideIndex` is this:
```julia
julia> A = zeros(3, 4, 5);

julia> A[:] = 1:60;

julia> Ap = @view(PermutedDimsArray(A,(3,1,2))[:,1:2,1])';

julia> ap_index = ArrayInterface.StrideIndex(Ap);

julia> ap_index[2,2] == 14 == Ap[2, 2]
true

julia> Ap[14]
ERROR: BoundsError: [...]

julia> parent(Ap)[14]
ERROR: BoundsError: [...]
```

We already automatically combine the strides all the way to the most deeply nested parent, but we don't have the proper semantics here to access that buffer properly (which is why it's not part of the indexing pipeline yet).
The point of the "layouts" PRs has been that we need a way to strip apart the index and buffer components at each layer of a nested array and then combine them into an optimized representation.

> IMO, unsafe_load(pointer(A)) == first(A), which is true for SubArrays because pointer (and memory_reference) offset the pointer when called so that this invariant holds.
>
> Under the perspective that the base pointer of a SubArray is the pointer to its first element (rather than the pointer to the parent), then offset1 currently does return the linear offset.
And for A::OffsetArray{T,N} where N > 1, we do not have that offset1(A) == first(offsets(A)).

I'm completely fine with that perspective, but it I'm not sure when we'd ever need a unique value for the `offset1` field of `StrideIndex`.
We'd just treat the method `offset1` like [`Base.axes1`](https://github.com/JuliaLang/julia/blob/0224c42bce6daec1b5580b8a8a6b43f9437d2cee/base/abstractarray.jl#L115) where it always redirects to the first dimension (except for zero dims).

> If these arrays are all strided, I'd rather have a single representation capable of describing any of them.

It's been a while, but in an old PR [we discussed](https://github.com/JuliaArrays/ArrayInterface.jl/pull/141#discussion_r627056938) how we needed to be able to account for non-strided indexing too. We don't need to solve every corner case, but we should probably provide standard `ArrayIndex` representations that can be combined in other packages.

I should probably just do a small PR that provides the few `ArrayIndex` subtypes I'm trying to describe right now.

> I am happy with the current semantics, I just wanted to elliminate redundant data storage.
LoopVectorization tries to pass data in registers instead of on the stack, but the number of registers you can use for this is very limited.

If there's a situation where we really need a unique `offset1` field then I think this PR looks good to merge, but if we don't need the field it might be simpler to just get rid of it entirely.

chriselrod:
> Neither `A[StrideIndex(A)[inds...]]` nor `parent(A)[StrideIndex(A)[inds...]` is technically the current way this functions.

Should've said `unsafe_load(pointer(A), StrideIndex(A)[inds...])`.
```julia
julia> using ArrayInterface

julia> A = zeros(3, 4, 5);

julia> A[:] = 1:60;

julia> Ap = @view(PermutedDimsArray(A,(3,1,2))[:,1:2,1])';

julia> ap_index = ArrayInterface.StrideIndex(Ap);

julia> ap_index[2,2] == 14 == Ap[2, 2]
true

julia> unsafe_load(pointer(Ap), ap_index[2,2])
14.0
```
Which is close to the behavior I actually care about.

"close", because of course I'm using `memory_reference` instead of `pointer`.

> I'm completely fine with that perspective, but it I'm not sure when we'd ever need a unique value for the offset1 field of StrideIndex.
We'd just treat the method offset1 like Base.axes1 where it always redirects to the first dimension (except for zero dims).

Mostly, it makes this easier to work with:
```julia
julia> using OffsetArrays

julia> A = OffsetArray(rand(3,4), 4, 4)
3×4 OffsetArray(::Matrix{Float64}, 5:7, 5:8) with eltype Float64 with indices 5:7×5:8:
 0.695104  0.0620855  0.374289  0.971638
 0.889623  0.765056   0.812169  0.820055
 0.693763  0.876049   0.084865  0.819834

julia> Base.axes1(A)
OffsetArrays.IdOffsetRange(values=5:7, indices=5:7)

julia> ArrayInterface.offset1(A)
static(1)

julia> ArrayInterface.offsets(A)
(5, 5)
```
Currently, the code for index canonicalization is quite awkward.
But, I wouldn't be opposed to just dropping `offset1`, because that awkward code works for now so no point replacing it.

chriselrod:
> It's been a while, but in an old PR [we discussed](https://github.com/JuliaArrays/ArrayInterface.jl/pull/141#discussion_r627056938) how we needed to be able to account for non-strided indexing too. We don't need to solve every corner case, but we should probably provide standard `ArrayIndex` representations that can be combined in other packages.
> 
> I should probably just do a small PR that provides the few `ArrayIndex` subtypes I'm trying to describe right now.

I think using `StrideIndex` should be preferable whenever possible, but using these wrapper types when not possible (e.g. permuting the dims or reshaping a banded matrix), that is fine.
For the permuting dims case though, I think we should just have every type represent their dimension ranks.

Maybe `stride_rank` is a bad word for it, but we could view it as a permutation on the indices whenever we do cartesian indexing.

Tokazama:
Looking over all of this I think it's clear I caused several issues with how I implemented `offset1` (the method and field of `StrideIndex`). I think the best path forward is to have `offset1(x)` == `offsets(x)[1]`. It's nice to have for vectors, but any other interpretation seems needlessly complicated at this point.

It would be nice to have `memory_reference` in ManualMemory.jl so that ArrayInterface.jl can depend on it, but other issues discussed here probably don't need to be resolved in this PR.

chriselrod:
> It would be nice to have `memory_reference` in ManualMemory.jl so that ArrayInterface.jl can depend on it, but other issues discussed here probably don't need to be resolved in this PR.

Seems reasonable. I'll move it.

I'm closing this PR in favor of  #190

chriselrod:
`memory_reference` depends on `ArrayInterface`, so it isn't straightforward to move.

PR #192, opened 2021-08-23
Assume parent array has same strides

Tokazama:
Assume strides are same for parent unless explicitly defined now. I'm not completely sure we are dealing with `ReshapedArray` in the best way, but I'm also not sure there is an ideal approach.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterface"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "3.1.24"
+version = "3.1.25"
 
 [deps]
 IfElse = "615f187c-cbe4-4ef1-ba3b-2fcf58d6d173"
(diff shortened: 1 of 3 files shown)


chriselrod:
Seems like `reinterpret` is hitting this for Julia 1.2.

chriselrod:
None of my libraries support Julia versions older than 1.5, so I'm fine with making the `reinterpret` tests conditional only trigger for these newer Julia versions.

Tokazama:
> None of my libraries support Julia versions older than 1.5, so I'm fine with making the reinterpret tests conditional only trigger for these newer Julia versions.

Is that going to be a sufficient solution or do you think it will break other packages? I'm assuming your stuff is the only thing directly using strides but I haven't done any formal downstream testing (nor do I know how to do that for over 700 packages that depend on this.

chriselrod (approved):
LGTM.

PR #195, opened 2021-08-28
`buffer` method

Tokazama:
See https://github.com/JuliaSIMD/ManualMemory.jl/pull/10 for context
Also added `parent_type(::Type{<:StaticArrays.SizedArray})` and `buffer` for `SArray` and `MArray`.

### DIFF
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -82,6 +82,16 @@ _has_parent(::Type{T}, ::Type{T}) where {T} = False()
 _has_parent(::Type{T1}, ::Type{T2}) where {T1,T2} = True()
 
 """
+    buffer(x)
+
+Return the buffer data that `x` points to. Unlike `parent(x::AbstractArray)`, `buffer(x)`
+may not return another array.type.
+"""
+buffer(x) = parent(x)
+buffer(x::SparseMatrixCSC) = getfield(x, :nzval)
+buffer(x::SparseVector) = getfield(x, :nzval)
+
+"""
     known_length(::Type{T}) -> Union{Int,Nothing}
 
 If `length` of an instance of type `T` is known at compile time, return it.
@@ -698,6 +708,9 @@ function __init__()
         can_setindex(::Type{<:StaticArrays.StaticArray}) = false
         ismutable(::Type{<:StaticArrays.MArray}) = true
         ismutable(::Type{<:StaticArrays.SizedArray}) = true
+
+        buffer(A::Union{StaticArrays.SArray,StaticArrays.MArray}) = getfield(A, :data)
+        parent_type(::Type{<:StaticArrays.SizedArray{<:Any,<:Any,<:Any,<:Any,P}}) where {P} = P
 
         function lu_instance(_A::StaticArrays.StaticMatrix{N,N}) where {N}
             A = StaticArrays.SArray(_A)
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterface"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "3.1.26"
+version = "3.1.27"
 
 [deps]
 IfElse = "615f187c-cbe4-4ef1-ba3b-2fcf58d6d173"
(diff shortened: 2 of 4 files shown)


chriselrod (approved):
Maybe we could have it in `ManualMemory`, but I think for now pairing it with the `ArrayIndex`es makes sense, since you probably don't want one without the other?

src/ArrayInterface.jl
```diff
@@ -81,6 +81,16 @@ has_parent(::Type{T}) where {T} = _has_parent(parent_type(T), T)
 _has_parent(::Type{T}, ::Type{T}) where {T} = False()
 _has_parent(::Type{T1}, ::Type{T2}) where {T1,T2} = True()
 
+"""
+    buffer(x)
+
+Return the buffer data that `x` points to. Unlike `parent(x::AbstractArray)`, `buffer(x)`
+may not return another array.type.
```
johnnychen94, line 88: Hmmm, what does this mean? `typeof(x)` and `typeof(buffer(x))` must be the same type?

Tokazama:
I've been trying to break things up into more manageable PRs, but it also make more overall objective a little more ambiguous. My plan is:

1. Next I'm going to add some subtypes `ArrayIndex` and how to combine them.
2. Finalize the "layouts" approach since everything will be in place at this point (something like `layout(A, access_pattern) = ArrayIndex(A), buffer(A)`).
3. Clean up "src/indexing.jl" while taking advantage of layouts.

I figured that 1 and 2 would set things up nicely for `LayoutPointers` and `VectorizationBase` to focus on optimization.

src/ArrayInterface.jl
Tokazama: If you're referring to `array.type`, it's a typo.

johnnychen94: It's not clear to me the terms "may not" and "another" here. Does "may not" mean "must not" or does it allow some sort of relaxation?

If `x` is a matrix, is it allowed for `buffer` to be a vector? If `x` is an OffsetArray, is it allowed for `buffer(x)` to be `Array`?

Or, as long as `buffer(x)` is a contiguous memory block whose element type is the same as `eltype(x)` it is a valid buffer?

Tokazama: > Or maybe here buffer doesn't mean buffer block, maybe it means something like "real storage".

More like this. `parent` is typically used on an array to return a wrapped array (not sure of an instance where it doesn't actually). If we want to be able to refer to the raw storage of values for arrays like `Tridiagonal` this criteria needs to be relaxed. There are also arrays like `SparseMatrixCSC` where the buffer is still an array type but isn't return with `parent`.

I made the explanation for what `buffer` returns here intentionally lax because we shouldn't need to dictate how data is stored to much; but perhaps it's just vague. We may want to add suggestions in the future (e.g., whatever is returned has `getindex` defined, supports `Ptr` or `ManualMemory.PseudoPtr`), but I don't see an immediate need for it and documenting that sort of requirement might force us into a breaking change when we encounter exceptions later on.

For `OffsetArray` nothing needs to happen because it already has `parent` defined, which returns its buffer.

johnnychen94: I get your points now. Can we rename this method to something more accurate? Maybe `raw_data`? Ask because I feel the name is perhaps ambiguous here with the purpose of "allocating a small array to serve as a buffer".

Tokazama: The reason I chose `buffer` is because we have `ManualMemory.preserve_buffer` which digs down to the deepest "buffer".
I think the return of `preserve_buffer` would be more accurately described as "raw data" because what I'm calling "buffer" here may have additional information attached to it, like another wrapper array.

> the name is perhaps ambiguous here with the purpose of "allocating a small array to serve as a buffer"

If we are looking at a method name for allocating array memory (outside of `similar), I think it would be odd to have a name that isn't derived from `malloc`/`alloc`.

All of this is clearly just my opinion so if there's a consensus for a clearly better name I'd be completely open to a rename.

Tokazama: I've waited a couple days to see if anyone else wanted to chime, but I'm gonna register the latest version now. I don't anticipate changing it being a breaking change for other packages within the immediate future (unless someone is closely following this repo and integrating every new method immediately). So if people want to discuss it more I'd encourage you to create an issue.

PR #197, opened 2021-08-30
Uint static step

chriselrod:
The added test fails on current master.

This causes problems for `StrideArraysCore` which uses unsigned `Base.OneTo`s to avoid the check.


chriselrod:
An alternative solution would be to add support for `static(::Uint64)`, e.g. by conversion or casting to `Int64`.

chriselrod:
Another solution would be to always use `StaticInt` instead of `static`, as this was already defined to convert.

Tokazama (commented):
LGTM

PR #199, opened 2021-09-01
ShapedIndex

Tokazama:
Primary goal here was to get the indexing pipeline closer to using `ArrayIndex` in the final steps (after index conversion/checking). Previously there was an awkward step after `unsafe_getindex` where linear -> cartesian (and vis versa) went back to `to_index`. Now `ShapedIndex` manages conversion to cartesian indices and `StrideIndex` is used for conversion to linear indexing (where the strides are computed using `size_to_strides`,  instead of the internal memory representation).

Getting this working also involved the following changes unrelated to to `ShapedIndex`:

* Previously an array that needed a unique method for `known_size(::Type{A})` also needed a unique method for `known_size(::Type{A}, dim)`. Now `known_size(::Type{A})` is called and then indexed, requiring only one new method.j
* Indexing `StrideIndex` was still using the old definition of `offset1`

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterface"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "3.2.0"
+version = "3.1.31"
 
 [deps]
 IfElse = "615f187c-cbe4-4ef1-ba3b-2fcf58d6d173"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterface"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "3.2"
+version = "3.2.0"
 
 [deps]
 IfElse = "615f187c-cbe4-4ef1-ba3b-2fcf58d6d173"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterface"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "3.1.30"
+version = "3.2"
 
 [deps]
 IfElse = "615f187c-cbe4-4ef1-ba3b-2fcf58d6d173"
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -81,6 +81,7 @@ ArrayInterface.BroadcastAxis
 ArrayInterface.LazyAxis
 ArrayInterface.OptionallyStaticStepRange
 ArrayInterface.OptionallyStaticUnitRange
+ArrayInterface.ShapedIndex
 ArrayInterface.StrideIndex
 ```
 
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -81,7 +81,6 @@ ArrayInterface.BroadcastAxis
 ArrayInterface.LazyAxis
 ArrayInterface.OptionallyStaticStepRange
 ArrayInterface.OptionallyStaticUnitRange
-ArrayInterface.ShapedIndex
 ArrayInterface.StrideIndex
 ```
 
--- a/src/array_index.jl
+++ b/src/array_index.jl
@@ -204,48 +204,11 @@ struct StrideIndex{N,R,C,S,O} <: ArrayIndex{N}
     end
 end
 
-"""
-    ShapedIndex(offsets::Tuple{Vararg{CanonicalInt}}}, size::Tuple{Vararg{CanonicalInt}}})
-
-Subtype of `ArrayIndex` that transforms a linear index into a multidimensional index.
-"""
-struct ShapedIndex{N,O,S} <: ArrayIndex{N}
-    offsets::O
-    size::S
-
-    function ShapedIndex(o::Tuple{Vararg{CanonicalInt,N}}, s::Tuple{Vararg{CanonicalInt,N}}) where {N}
-        new{N,typeof(o),typeof(s)}(o, s)
-    end
-    ShapedIndex(x) = ShapedIndex(offsets(x), size(x))
-end
-
 Base.firstindex(i::Union{TridiagonalIndex,BandedBlockBandedMatrixIndex,BandedMatrixIndex,BidiagonalIndex,BlockBandedMatrixIndex}) = 1
 Base.lastindex(i::Union{TridiagonalIndex,BandedBlockBandedMatrixIndex,BandedMatrixIndex,BidiagonalIndex,BlockBandedMatrixIndex}) = i.count
 Base.length(i::Union{TridiagonalIndex,BandedBlockBandedMatrixIndex,BandedMatrixIndex,BidiagonalIndex,BlockBandedMatrixIndex}) = i.count
 
 ## getindex
-Base.getindex(x::ShapedIndex, i::AbstractCartesianIndex) = i
-Base.getindex(x::ShapedIndex, i::CanonicalInt) = _lin2sub(offsets(x), size(x), i)
-@generated function _lin2sub(o::O, s::S, i::I) where {O,S,I}
-    out = Expr(:block, Expr(:meta, :inline))
-    t = Expr(:tuple)
-    iprev = :(i - 1)
-    N = length(S.parameters)
-    for i in 1:N
-        if i === N
-            push!(t.args, :($iprev + getfield(o, $i)))
-        else
-            len = gensym()
-            inext = gensym()
-            push!(out.args, :($len = getfield(s, $i)))
-            push!(out.args, :($inext = div($iprev, $len)))
-            push!(t.args, :($iprev - $len * $inext + getfield(o, $i)))
-            iprev = inext
-        end
-    end
-    push!(out.args, :(NDIndex($(t))))
-    out
-end
 @propagate_inbounds Base.getindex(x::ArrayIndex, i::CanonicalInt, ii::CanonicalInt...) = x[NDIndex(i, ii...)]
 @propagate_inbounds function Base.getindex(ind::BidiagonalIndex, i::Int)
     @boundscheck 1 <= i <= ind.count || throw(BoundsError(ind, i))
--- a/src/array_index.jl
+++ b/src/array_index.jl
@@ -205,7 +205,7 @@ struct StrideIndex{N,R,C,S,O} <: ArrayIndex{N}
 end
 
 """
-    ShapedIndex(size::Tuple{Vararg{CanonicalInt}}})
+    ShapedIndex(offsets::Tuple{Vararg{CanonicalInt}}}, size::Tuple{Vararg{CanonicalInt}}})
 
 Subtype of `ArrayIndex` that transforms a linear index into a multidimensional index.
 """
--- a/src/array_index.jl
+++ b/src/array_index.jl
@@ -204,9 +204,48 @@ struct StrideIndex{N,R,C,S,O} <: ArrayIndex{N}
     end
 end
 
+"""
+    ShapedIndex(size::Tuple{Vararg{CanonicalInt}}})
+
+Subtype of `ArrayIndex` that transforms a linear index into a multidimensional index.
+"""
+struct ShapedIndex{N,O,S} <: ArrayIndex{N}
+    offsets::O
+    size::S
+
+    function ShapedIndex(o::Tuple{Vararg{CanonicalInt,N}}, s::Tuple{Vararg{CanonicalInt,N}}) where {N}
+        new{N,typeof(o),typeof(s)}(o, s)
+    end
+    ShapedIndex(x) = ShapedIndex(offsets(x), size(x))
+end
+
 Base.firstindex(i::Union{TridiagonalIndex,BandedBlockBandedMatrixIndex,BandedMatrixIndex,BidiagonalIndex,BlockBandedMatrixIndex}) = 1
 Base.lastindex(i::Union{TridiagonalIndex,BandedBlockBandedMatrixIndex,BandedMatrixIndex,BidiagonalIndex,BlockBandedMatrixIndex}) = i.count
 Base.length(i::Union{TridiagonalIndex,BandedBlockBandedMatrixIndex,BandedMatrixIndex,BidiagonalIndex,BlockBandedMatrixIndex}) = i.count
+
+## getindex
+Base.getindex(x::ShapedIndex, i::AbstractCartesianIndex) = i
+Base.getindex(x::ShapedIndex, i::CanonicalInt) = _lin2sub(offsets(x), size(x), i)
+@generated function _lin2sub(o::O, s::S, i::I) where {O,S,I}
+    out = Expr(:block, Expr(:meta, :inline))
+    t = Expr(:tuple)
+    iprev = :(i - 1)
+    N = length(S.parameters)
+    for i in 1:N
+        if i === N
+            push!(t.args, :($iprev + getfield(o, $i)))
+        else
+            len = gensym()
+            inext = gensym()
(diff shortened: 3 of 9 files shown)


Tokazama:
It looks like the change in test coverage is mostly do to the deletion of code that was previously being tested but is now redundant. It will probably work better for me to address that in conjunction with my next PR.

timholy:
How about just
```julia
ci = reshape(CartesianIndices(A), :)
ci[i]   # gives the CartesianIndex corresponding to i::Int
```
That's approximately as fast as a single `div` and you get all of them.

chriselrod:
`_lin2sub` is using `div` so it won't be faster than a `div` instruction, so it isn't doing any magic to be faster.
Any reason not to implement it using `reshape(CartesianIndices(A), :)`?

Tokazama:
The most obvious benefit here is the ability to incorporate `StaticInt`, which gives a very small speed boost. 

> and you get all of them

In this case we don't want a representation that will stick around and represents an entire array. The long game here is that it provides a minimalistic representation of an index transform that we can use in the future when creating rules for combining nested index transforms.

timholy:
> In this case we don't want a representation that will stick around and represents an entire array.

```julia
julia> sizeof(ci)
48

julia> ci.
dims   mi      parent
julia> ci.dims
(2000000,)

julia> ci.mi
(Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}(1000, 2361183241434822607, 0, 0x07),)

julia> cii = ci.parent;

julia> typeof(cii)
CartesianIndices{2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}

julia> cii.indices
(Base.OneTo(1000), Base.OneTo(2000))
```
There's "nothing" there. What you really want are those `SignedMultiplicativeInverse`s.

And it's even better than I suggested, as much of the time in `ci[i]` is bounds-checking. Intrinsically it's faster than a single `div`.

Tokazama:
I'm fine with working in `reshape(CartesianIndices(A), :)`, but I'm not sure where performance will benefit from that approach. Maybe I'm just benchmarking this incorrectly.

```julia
julia> to_cartesian_reshaped_array(x, i) = @inbounds(reshape(CartesianIndices(axes(x)), :)[i]);

julia> to_cartesian_shaped_index(x, i) = ArrayInterface.ShapedIndex(x)[i];

julia> A = @SArray(zeros(2,2,2));

julia> @btime to_cartesian_reshaped_array($A, 2)
  9.127 ns (0 allocations: 0 bytes)
CartesianIndex(2, 1, 1)

julia> @btime to_cartesian_shaped_index($A, 2)
  0.042 ns (0 allocations: 0 bytes)
NDIndex(2, 1, 1)

```

timholy:
Anytime you get < 1 clock cycle (≈1/3ns) you know the compiler is fooling you. Use `setup`. Also, can you not precalculate the reshaped array? (i.e., hoist it out of a loop.) It invests time at the outset to make it all much faster later.

chriselrod:
I get
```julia
julia> @btime to_cartesian_shaped_index($A, $(Ref(2))[])
  1.644 ns (0 allocations: 0 bytes)
NDIndex(2, 1, 1)

julia> @btime to_cartesian_reshaped_array($A, $(Ref(2))[])
  7.193 ns (0 allocations: 0 bytes)
CartesianIndex(2, 1, 1)
```
The `@code_typed` and `@code_native` are much nicer for `to_cartesian_shaped_index`.
FWIW, [uiCA says `to_cartesian_shaped_index` would take 6.33 clock cycles on average on my CPU (TigerLake) in unrolled code](https://bit.ly/3gRg0T4). If the CPU were running at 4 GHz, it could achieve a 1.644 ns time.
Obviously <0.1 ns is impossible (a clock cycle at 4 GHz takes 0.25 ns), but using `$(Ref(2))[]` above seemed to give a reasonable time.

Because `A` is a `StaticArray`, the compiler will of course normally fully specialize on it.

The `to_cartesian_reshaped_array` has a lot of branches, which aren't supported by uiCA, so I can't get a similar estimate there.
Despite the `@inbounds`, there are three bounds errors, plus one inexacterror getting called in the assembly (behind branches).

johnnychen94:
I didn't observe any big performance difference here. The `reshape(, :)` is actually unnecessary because it will use linear indexing here.

```julia
julia> function sumcart(X)
           out = zero(eltype(X))
           R = CartesianIndices(X)
           @inbounds @simd for i in eachindex(X)
               out += X[R[i]]
           end
           return out
       end
sumcart (generic function with 1 method)

julia> function sumcart_shaped(X)
           out = zero(eltype(X))
           R = ArrayInterface.ShapedIndex(X)
           @inbounds @simd for i in eachindex(X)
               out += X[R[i]]
           end
           return out
       end
sumcart_shaped (generic function with 1 method)

julia> @btime sumcart($SX);
  3.828 ns (0 allocations: 0 bytes)

julia> @btime sumcart_shaped($SX);
  4.231 ns (0 allocations: 0 bytes)

julia> X = rand(64, 64);

julia> @btime sumcart($X);
  230.455 ns (0 allocations: 0 bytes)

julia> @btime sumcart_shaped($X);
  229.612 ns (0 allocations: 0 bytes)
```

Tokazama:
@timholy, is this what you wanted?

```julia
julia> @benchmark getindex(A, 2) setup=(A=reshape(CartesianIndices(axes(@SArray(zeros(2,2,2)))), :))
BenchmarkTools.Trial: 10000 samples with 1000 evaluations.
 Range (min … max):  4.439 ns … 44.816 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     5.220 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   5.202 ns ±  1.342 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▂  ▃   ▃   ▂▁   ▃   ▁▃   ▅▇    ▆▂   ▂▃    ▃▅     ▄     ▂▆ ▂
  ██▁▁█▇▁▁█▁▁▁██▁▃▇█▁▁▁██▄▁▁███▁▃▄██▁▁▁██▆▁▁▁██▆▁▁▁▁█▇▁▁▁▁██ █
  4.44 ns      Histogram: log(frequency) by time      6.1 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark getindex(A, 2) setup=(A=ArrayInterface.ShapedIndex(@SArray(zeros(2,2,2))))
BenchmarkTools.Trial: 10000 samples with 1000 evaluations.
 Range (min … max):  0.040 ns … 0.112 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     0.046 ns             ┊ GC (median):    0.00%
 Time  (mean ± σ):   0.046 ns ± 0.002 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                     ▂      █      ▅
  ▂▁▁▁▁▁▁▂▁▁▁▁▁▁▃▁▁▁▁▁▁▅▁▁▁▁▁▁▆▁▁▁▁▁▁█▁▁▁▁▁▁█▁▁▁▁▁▁█▁▁▁▁▁▁▃ ▂
  0.04 ns        Histogram: frequency by time      0.048 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.

```

chriselrod:
Odd, I see
```julia
julia> X = rand(64, 64);

julia> @btime sumcart($X)
  5.362 μs (0 allocations: 0 bytes)
2044.2126606418335

julia> @btime sumcart_shaped($X)
  115.387 ns (0 allocations: 0 bytes)
2044.2126606418326
```

chriselrod:
> is this what you wanted?

The `2` is what it isn't allowed to specialize on.
It only cares about the type of `A::StaticArray`, so randomly creating different ones shouldn't change the outcome.

johnnychen94:
> Odd, I see

Looks like it's messed up with bounds check. Sorry I didn't mention that I opened julia with `--check-bounds=no`

Tokazama:
> Also, can you not precalculate the reshaped array? (i.e., hoist it out of a loop.) It invests time at the outset to make it all much faster later.

Ultimately we want to consider doing this before the loop, but I figured indexing collections would be a dedicated PR later (once I've got more of the other indexing stuff in place).

chriselrod:
> > Odd, I see
> 
> Looks like it's messed up with bounds check. Sorry I didn't mention that I opened julia with `--check-bounds=no`

I can reproduce with `--check-bounds=no`.
Seems like it's missing a few `Base.@propagate_inbounds`.

Tokazama:
I get similar disparities in performance on an `Array`
```julia

julia> @btime sumcart_shaped($A)
  235.834 ns (0 allocations: 0 bytes)
2040.2091413638263

julia> @btime sumcart($A)
  10.060 μs (0 allocations: 0 bytes)
2040.2091413638286
```

Tokazama:
I'm not sure how to compare assembly. No matter what `@code_*` I use, it's just too many lines to eyeball a comparison for me.

timholy:
With StaticArray, LLVM essentially computes the `SignedMultiplicativeInverse` for you (at compile time, so it is cost-free). With a dynamic one, it cannot. https://en.wikipedia.org/wiki/Montgomery_modular_multiplication. So you can use the scheme you're proposing for Static arrays, but `div` is pretty much a disaster on some CPUs when you actually have to do integer division: http://ithare.com/infographics-operation-costs-in-cpu-clock-cycles/ (it's almost as bad as a function call). That said, that might be outdated, I have the sense that recent CPUs are better at it. (@chriselrod?)

timholy:
@johnnychen94,

> I didn't observe any big performance difference here

```julia
julia> @btime sumcart_reshape($X);
  105.498 μs (0 allocations: 0 bytes)

julia> @btime sumcart($X);
  160.048 μs (0 allocations: 0 bytes)
```

BUT it's different for tiny arrays (this was 200x300, dynamic not static since the compiler does the same thing as `reshape` for static). And it's quite possible that the `@simd` may have trouble reasoning about the integers...

chriselrod:
> That said, that might be outdated, I have the sense that recent CPUs are better at it. (@chriselrod?)

It's gotten a lot better in recent CPUs, but is still bad. [64 bit div](https://uops.info/html-instr/DIV_R64.html)
Cycles per div
| Zen(+/2) | Zen3 | Intel Skylake and older | Ice/Tiger/Rocket Lake |
|----|----|---|----|
| 14 | 7 | 21 | 10 |

So Zen3 CPUs and Intel Ice/Tiger/Rocket Lake (newest from AMD and Intel respectively) are both faster than reported in that chart (15-40), but these are still slow.
Every 7 clock cycles, Zen3 can do 14x 256 bit FMAs.

Tokazama:
I love that this summoned the benchmarking A-team.

Just a quick example where this will be beneficial in the future.
In a future PR I'm going to add another `ArrayIndex` subtype that accounts for dimension permutations.
If we have `PermutedDimsArray(PermutedDimsArray(A, (3,1,2)), (2, 3, 1))` then they negate each other, but a linear index would still undergo an initial transformation to a cartesian index.
If we flatten out the index transformations into simple representations and combine them, then we don't return something that uses `ShapedIndex` and can avoid related method calls at run time.

timholy:
> Is there a good api on the signed multiplicative inverses?

They were added in https://github.com/JuliaLang/julia/pull/15357, and the code is https://github.com/JuliaLang/julia/blob/master/base/multinverses.jl. That's pretty much all there is.

timholy:
> PermutedDimsArray(PermutedDimsArray(A, (3,1,2)), (2, 3, 1))

Examples like these are what made me initially add `permuteddimsview` to ImageCore. IMO `PermutedDimsArray` is a constructor and should build one, but `permuteddimsview` is a function and can do whatever it wants. Obviously that's (mostly) just a convention, but I've found it becomes important for inferrability in cases of recursion.

Tokazama:
> IMO PermutedDimsArray is a constructor and should build one, but permuteddimsview is a function and can do whatever it wants.

I completely agree with this. The focus with subtypes of `ArrayIndex` is to have a set of concise, flexible, and composable index transformations. Some of them may only require a buffer to compose a complete array, but others will require additional context (for example, any strided array could be represented by a buffer, tuple of sizes, and a `StrideIndex`). So if we do end up using these to construct arrays (a la https://github.com/JuliaLang/julia/issues/18161), we'll need to figure out how an `ArrayIndex` interacts with other arguments to construct a valid array.

Tokazama:
Just to summarize:

On my 2019 macbook I get the following on 1.7 with bounds checking off.

```julia
julia> function sum_cartesian(X)
           out = zero(eltype(X))
           R = CartesianIndices(X)
           @inbounds @simd for i in eachindex(X)
               out += X[R[i]]
           end
           return out
       end
sum_cartesian (generic function with 1 method)

julia> function sum_reshaped(X)
           out = zero(eltype(X))
           R = CartesianIndices(X)[:]
           @inbounds @simd for i in eachindex(X)
               out += X[R[i]]
           end
           return out
       end
sum_reshaped (generic function with 1 method)

julia> function sum_shaped(X)
           out = zero(eltype(X))
           R = ArrayInterface.ShapedIndex(X)
           @inbounds @simd for i in eachindex(X)
               out += X[R[i]]
           end
           return out
       end
sum_shaped (generic function with 1 method)

julia> SA = @SArray(rand(4, 4));

julia> A=rand(64, 64);

julia> @btime sum_cartesian($A)
  229.594 ns (0 allocations: 0 bytes)
2054.4112898827025

julia> @btime sum_reshaped($A)
  9.409 μs (2 allocations: 64.05 KiB)
2054.4112898827025

julia> @btime sum_shaped($A)
  229.171 ns (0 allocations: 0 bytes)
2054.4112898827025

julia> @btime sum_cartesian($SA)
  3.312 ns (0 allocations: 0 bytes)
7.888377370714911

julia> @btime sum_reshaped($SA)
  65.352 ns (1 allocation: 336 bytes)
7.888377370714911

julia> @btime sum_shaped($SA)
  2.958 ns (0 allocations: 0 bytes)
7.888377370714911
```

Differences in performance of `ShapedIndex` and `CartesianIndices` are marginal once bounds checking is off.
I think there is some merit in having `ShapedIndex` being a subtype of `ArrayIndex` and being slightly simpler than `CartesianIndices`, but if we get the same performance putting a PR into base for the missing `@propagate_inbounds` then we could just define a constant that uses `OptionallyStaticUnitRange` to ensure we have essentially the same thing when dispatching.

It sounds like there are some benefits to `SignedMultiplicativeInverse`, but they are going to require optimizing with hardware in mind.
This makes me think something like `LoopVectorization` should be managing it.
If we need to define a simple method or type that's commonly available that might be worth considering, but I don't have any ideas at the moment on the best way to implement that, so perhaps best to approach in another PR?

chriselrod:
```julia
R = CartesianIndices(X)[:]
```
That should be
```julia
R = rehape(CartesianIndices(X), :)
```
to avoid allocating a `Vector`.

Tokazama:
Sorry, I fixed updated it.

Tokazama:
K. I dropped `ShapedIndex` in favor of `CartesianIndices` composed through a call to `ArrayInterface.indices`

Tokazama:
@chriselrod , I moved to a patch bump here so the bug fixes here can go forward.

chriselrod (commented):
Sure, this is fine if you just want to get this in. Without `@propagate_inbounds` in the correct places in base, the implementation without cartesian indices will probably often be faster in practice.

chriselrod (approved):

Tokazama:
> Without @propagate_inbounds in the correct places in base, the implementation without cartesian indices will probably often be faster in practice.

I'll look into making a PR to base. If it turns out to be insurmountable, then we have a solid conversation here and documented effort that shows `ShapedIndex` wouldn't be a redundant addition.

PR #204, opened 2021-09-10
adapt to JuliaLang/julia#42125

simeonschaub:


mcabbott:
There is also https://github.com/JuliaLang/Compat.jl/pull/752 btw. Which is a more elaborate one allowing for the intermediate versions like like 1.7-beta. But whether 800-odd packages want to suddenly depend on Compat.jl I don't know.

Tokazama:
Yeah, Compat is a bit of a double edged sword. I think we're hoping to do easy fixes on our own until we get the next LTS of Julia.

mcabbott:
What I meant to say, but see I didn't, is that this could steal that PR's implementation, if this package wants to keep working on 1.7-beta versions. Or else the `VERSION >= v"1.7.0-DEV.421"` bound may need to change?

simeonschaub:
> What I meant to say, but see I didn't, is that this could steal that PR's implementation, if this package wants to keep working on 1.7-beta versions.

Since that PR will be backported, I don´t think this is really worth it.



> Or else the `VERSION >= v"1.7.0-DEV.421"` bound may need to change?

I think it might be smarter to instead check for `isdefined(Base, Symbol("@constprop"))` here. Should I do that for Static.jl as well?

simeonschaub:
Looks like someone still needs to approve this run (I think you can also change that as an org-wide setting). Otherwise this should be good to go.

Project.toml
```diff
@@ -12,7 +12,7 @@ Static = "aedffcd0-7271-4cad-89d0-dc628f76c6d3"
 [compat]
 IfElse = "0.1"
 Requires = "0.5, 1.0"
-Static = "0.3"
+Static = "0.3.2"
```
simeonschaub, line 15:
```suggestion
Static = "0.3.3"
```
(Once it's tagged)

chriselrod (approved):
Thanks for all the updates!
Sorry I accidentally nuked Julia 1.7.0-4beta and wasn't more responsive. I was out of office/on vacation with family.

PR #205, opened 2021-09-11
Use `Compat.@constprop` and `Compat.@inline`

timholy:
These are changing names/gaining functionality.
Compat allows one to easily support multiple Julia versions.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,8 +1,9 @@
 name = "ArrayInterface"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "3.1.32"
+version = "3.1.33"
 
 [deps]
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 IfElse = "615f187c-cbe4-4ef1-ba3b-2fcf58d6d173"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
@@ -10,6 +11,7 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Static = "aedffcd0-7271-4cad-89d0-dc628f76c6d3"
 
 [compat]
+Compat = "3.37"
 IfElse = "0.1"
 Requires = "0.5, 1.0"
 Static = "0.3.3"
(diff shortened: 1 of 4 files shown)


timholy:
CC @aviatesk

timholy:
Ah, I thought I was editing my own previous change, now I see this essentially happened in #204. This can be closed unless you want the `@inline` (which fixes a depwarn).

aviatesk (approved):
yeah, I still prefer this PR because the `Compat` version does additional argument validation etc., so I think it's reasonable to merge it if we're okay to introduce new dependency on Compat.jl.

Tokazama:
I don't have a strong opinion on this one, but if it's a better band-aid it might be worth further discussion. Is the the main concern here the extra Compat.jl dependency? If so, could we just negotiate this concern by agreeing to drop Compat.jl with the next LTS? I'm assuming it will come around in the next yearish.

timholy:
Yes, if this package will drop old Julia releases you can handle things that way. Compat's size as a dependency also decreases for people running newer Julia versions:
```sh
tim@diva:~/.julia/compiled$ ls -l v1.0/Compat/*.ji
-rw------- 1 tim holy 226711 Aug 31 04:33 v1.0/Compat/GSFWK.ji
tim@diva:~/.julia/compiled$ ls -l v1.8/Compat/*.ji
-rw-r--r-- 1 tim holy 69209 Aug 28 19:36 v1.8/Compat/GSFWK_eYLmO.ji
-rw-r--r-- 1 tim holy 69209 Sep  2 03:49 v1.8/Compat/GSFWK_FJfGL.ji
-rw-r--r-- 1 tim holy 70749 Sep 10 06:59 v1.8/Compat/GSFWK_mdmZ9.ji
-rw-r--r-- 1 tim holy 48135 Sep  1 05:09 v1.8/Compat/GSFWK_Ol4t2.ji
-rw-r--r-- 1 tim holy 72747 Sep 15 09:33 v1.8/Compat/GSFWK_V8kJc.ji
-rw-r--r-- 1 tim holy 70701 Sep 10 06:51 v1.8/Compat/GSFWK_wxM3A.ji
-rw-r--r-- 1 tim holy 70010 Sep 14 02:53 v1.8/Compat/GSFWK_yRJjT.ji
-rw-r--r-- 1 tim holy 69194 Aug 27 03:46 v1.8/Compat/GSFWK_z7wL7.ji
-rw-r--r-- 1 tim holy 69209 Aug 28 09:36 v1.8/Compat/GSFWK_ZcdGt.ji
```
So about 1/3. Most of that is probably defining `@compat` (which you then basically don't use).

chriselrod:
Compat may be [ubiquitous enough already](https://juliahub.com/ui/Packages/Compat/GSFWK/3.37.0?t=2) that most people are already pulling it in anyway.

Tokazama:
Sounds like everyone is fine with this as a dep. We just need a patch bump.

PR #210, opened 2021-09-26
ArrayInterface.axes for ReshapedReinterpretArray

timholy:
This specializes `axes` for ReshapedReinterpretArray, and streamlines a
bit of the code surrounding this type.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -433,7 +433,7 @@ ArrayInterface.parent_type(::Type{DenseWrapper{T,N,P}}) where {T,N,P} = P
     Am = @MMatrix rand(2,10);
     @test @inferred(ArrayInterface.strides(view(Am,1,:))) === (StaticInt(2),)
 
-    if VERSION ≥ v"1.6.0-DEV.1581" # reinterpret(reshape,...) tests
+    if isdefined(Base, :ReshapedReinterpretArray) # reinterpret(reshape,...) tests
         C1 = reinterpret(reshape, Float64, PermutedDimsArray(Array{Complex{Float64}}(undef, 3,4,5), (2,1,3)));
         C2 = reinterpret(reshape, Complex{Float64}, PermutedDimsArray(view(A,1:2,:,:), (1,3,2)));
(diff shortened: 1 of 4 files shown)


Tokazama:
Is this ready for review

chriselrod (approved):
I guess this library has transitioned from tabs = 2 -> tabs = 4 spaces?

timholy:
> Is this ready for review

Yes. It doesn't implement anything for non-reshaped ReinterpretArray (in a sense, that's the harder case), and I added a `@broken` test to indicate one unfortunate issue. But I think the implementation for a `reinterpret(reshape, ...)` is "done."

> I guess this library has transitioned from tabs = 2 -> tabs = 4 spaces?

Not sure if you mean they were tabs before and spaces now (sorry, if I didn't notice and perhaps I should have updated my definition of tab), or if just the default indent had changed. Mostly I found the mix awkward and the 4-space indent dominates, so I went with that.

test/runtests.jl
```diff
@@ -717,6 +717,8 @@ end
     @test @inferred(ArrayInterface.strides(u_view)) == (4,)
     @test @inferred(ArrayInterface.strides(u_view_reinterpreted)) == (4,)
     @test @inferred(ArrayInterface.strides(u_view_reshaped)) == (4, 4)
+
+    @test_broken @inferred(ArrayInterface.axes(u_vectors)) isa ArrayInterface.axes_types(u_vectors)
```
chriselrod, line 721: Does inference fail here?

timholy: No, it's that it isn't of the type returned by `axes_types`.

Tokazama: I need to do some clean up on `axes` and can fix this. I think we need to just go all in on `OptionallyStaticUnitRange` here.

timholy: No worries, and no hurry (there's nothing wrong with a broken test, it's essentially a TODO list).

Tokazama: Did you want me to fix this before this PR went forward or afterwards. Just want to make sure this PR isn't stalled because of me.

timholy:
Just checking, this effectively got merged in #222, right? So close?

Tokazama:
It incorporates the same tests so unless there was something else needed here this should be good to close

PR #222, opened 2021-10-27
`ArrayInterface.axes` docs, clean up, and `ReinterpretArray` support.

Tokazama:
* Get rid of some no good very bad spaghetti code.
* Document examples of how to use `axes` here.
* Incorporate `ReinterpretArray` changes from #210.
* Don't assume we can go to parent arrays anymore because there are too many assumptions (offsets, sizing, views, etc) that can change it.

@timholy, I copied the `ReinterpretArray` axes tests directly directly from #210 (except we no longer have a broken test). Was there anything else that you wanted covered here?

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -192,6 +192,7 @@ can_change_size(::Type{LazyAxis{N,P}}) where {N,P} = can_change_size(P)
 
 known_first(::Type{LazyAxis{N,P}}) where {N,P} = known_offsets(P, static(N))
 known_first(::Type{LazyAxis{:,P}}) where {P} = 1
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
This seems to have caused downstream breakage.

https://github.com/SciML/SciMLBase.jl/runs/4049107830?check_suite_focus=true#step:6:1030

https://github.com/SciML/DiffEqFlux.jl/issues/646

chriselrod:
The breakage was making `_axes` signature more specific, resulting in ambiguities.

ChrisRackauckas:
Yup just saw that in the other issue.

PR #227, opened 2021-11-08
Optimizations on to_indices

Tokazama:
This is a performance optimized version of `to_indices`. I've provided the justification for most changes from `Base.to_indices` in the "# Extended help" section of the doc string. It also includes some minimal benchmarking to drive the point home. I might change a few minor bits of code once I get some sleep, but it can work for any of the indexing argument types in currently registered packages and can do so faster than `Base.to_indices` in all cases (as far as I'm aware). I'm sure there are parts I haven't documented or explained well, but I think it's a pretty decent start for now.

The biggest hang ups I anticipate are potentially breaking changes (still need to check if we can take away the 3 arg version of `to_indices`) and the two large `@generated` methods working behind the scenes (`_splat_indices` and `__to_indices`). I think we can solve any problems that might come up with regards to accommodating new indexing types using traits.


Tokazama:
@ChrisRackauckas and/or @chriselrod, I know the inner `@generated` code is a bit messy for a thorough review here, but is the overall approach and motivation for the unique approach clear at all through the docs?

If we're worried about breaking changes with this, then we should probably address some additional breaking changes before releasing 4.0 (like #228).

chriselrod (approved):
I don't have time for a thorough review, but this looks fine at a glance.

PR #229, opened 2021-12-22
fix for `firstindex(128:static(-1):1)`

N5N3:
make `firstindex(128:static(-1):1)` return `1` instead of `128`


chriselrod (approved):
Mind bumping the version?

N5N3:
The error seems unrelated?

chriselrod:
Yes, they're also failing on master.

PR #232, opened 2021-12-30
Use `missing` instead of `nothing` for dynamic values

Tokazama:


chriselrod (approved):

Tokazama:
For some reason all tests are failing but at different places for different Julia versions. I'll have to do some more poking around.

chriselrod:
> For some reason all tests are failing but at different places for different Julia versions. I'll have to do some more poking around.

This fixes some of them: https://github.com/JuliaArrays/ArrayInterface.jl/pull/230

chriselrod:
Odd that this is getting an ambiguity error on `lu` for StaticArrays, when master isn't, and that code hasn't been touched in this PR?

Tokazama:
I assumed that `missing` would have similar behavior as `nothing` since they're both singletons, but maybe there's some sort of compiler wizardry going on with `nothing` I'm not aware of.

chriselrod:
The lu ambiguities are because it was using an old version of StaticArrays, before:
https://github.com/JuliaArrays/StaticArrays.jl/commit/7dc649948b26fc7e5c7525560d41064f221244a5

We could add a test/Project.toml to restrict the version it tests with.

chriselrod:
Doc tests fail, but should pass once the pkg server updates and has Static 0.5 available.
So this should be good to go.

Tokazama:
Since this is breaking, do we want to wait to register this until we manage other breaking changes like moving BandedMatrixIndex out

chriselrod:
> Since this is breaking, do we want to wait to register this until we manage other breaking changes like moving BandedMatrixIndex out

Sure, no rush.

Tokazama:
I just moved the banded matrix stuff into requires so that when we move to other packages nothing should change.

YingboMa:
Is this breaking change really worth to make? The whole SciML ecosystem is still forced on v3 and not able to use new features.

ChrisRackauckas:
It's been too long without downstream PRs. Let's revert it today.

Tokazama:
Is the problem just with getting LoopVectorization updated or do we need to go through the whole ecosystem? I'd be surprised if many packages had to change anything to accommodate this.

PR #233, opened 2021-12-31
parent(::LazyAxis) returns the axis.

chriselrod:
@Tokazama lets get this merged, tag a new release, and then we can bump for the static update.

The reason tests started failing on Julia >= 1.7 is because `size` switched from using `unsafe_length` in <= 1.6 -- which gave the correct answer -- to using `length` in >= 1.7 -- which didn't.


Tokazama (approved):
LGTM

PR #236, opened 2022-01-11
Support static info on iterators

Tokazama:
* New support for types from `Base.Iterators`
* Check for `Base.IteratorSize` before assuming we can rely on axes when
  finding the size.
* Update internal calculation of range length to match base

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterface"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "4"
+version = "4.0.1"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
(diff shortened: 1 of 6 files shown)


chriselrod (approved):
People have consistently found `indices` confusing, although I like it.
So it's probably nice to support more methods.
Also, not redundant, as this gives us sizes instead of an axis.

Tokazama:
> People have consistently found `indices` confusing, although I like it. So it's probably nice to support more methods. Also, not redundant, as this gives us sizes instead of an axis.

I can also add `indices` and `ArrayInterface.axes` support for iterators. I just didn't immediately need it so I didn't think of that

chriselrod:
> > People have consistently found `indices` confusing, although I like it. So it's probably nice to support more methods. Also, not redundant, as this gives us sizes instead of an axis.
> 
> I can also add `indices` and `ArrayInterface.axes` support for iterators. I just didn't immediately need it so I didn't think of that

No need. I was just commenting on the `size(zip(...))` tests.

PR #238, opened 2022-01-28
Dynamic dimnames

Tokazama:
The utility of dynamic dimension names came up a while ago in a slack conversion. I usually work with a very consistent set of dimension names that doesn't ever change, but packages that use `NamedArrays.jl` apply dimension names with a lot more variability (like different statistical analyses).

This implementation adds `known_dimnames` so that `dimnames` acts more like `size`/`known_size`, `offsets`/`known_offsets`. I've added a little bit to the "index.md" file explaining this. This also fixes a bug where an error was erroneously thrown when indexing by named slices.

I should mention that despite some extra work on catering to variable levels inferrability, dynamic dimensions will still have some issues with inference. For example, the last test in "test/dimensions.jl":

```julia
y = NamedDimsWrapper((:x, static(:y)), ones(2, 2));
getindex(y, x = 1) == [1, 1]
```

That last call gives up and infers `Any`.
If you dig into it, parsing the indices infers `Tuple{Union{Colon, Int64}, Union{Colon, Int64}}`, so we know logically the only outcome is a vector.
With some extra work I'm sure I could get this exact situation to infer, but I'm not sure how realistic it is that scales to more complex situations.
I'm keeping the parsing stuff strictly internal so that there's no breaking changes in the future if theres a fix for this that uses a slightly different approach.

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -26,6 +26,8 @@ jobs:
           - {user: SciML, repo: OrdinaryDiffEq.jl, group: Core}
           - {user: SciML, repo: OrdinaryDiffEq.jl, group: Interface}
           - {user: SciML, repo: DelayDiffEq.jl, group: Interface}
+          - {user: JuliaSIMD, repo: VectorizationBase.jl, group: Interface}
+          - {user: JuliaSIMD, repo: LoopVectorization.jl, group: Downstream3}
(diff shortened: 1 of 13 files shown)


Tokazama:
I'm going to go ahead and merge this since there's no real risk of it breaking downstream code ATM.

PR #241, opened 2022-02-08
ArrayInterface.Size instead of ArrayInterface.size

Tokazama:
This implements a formal type for extracting optionally static size info as I previously brought up [here](https://github.com/JuliaArrays/ArrayInterface.jl/issues/182#issuecomment-1005730557). If we do end up going this direction we can just leave a fall back to `ArrayInterface.size(x) = Size(x).size` so that we don't break any downstream code.

I think this is more beneficial than just creating a greater distinction between `Base.size` and `ArrayInterface.size`. It also allows passing `Size` as an unambiguous argument to methods like `similar` allocate a new array but currently dispatch on various combinations of integers and unit ranges for the `dims` argument.  I'm hoping this could be used in combination with `ArrayInterface.AbstractDevice` to give us a more complete story for array constructors.

### DIFF
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -572,8 +572,8 @@ end
 
 abstract type AbstractArray2{T,N} <: AbstractArray{T,N} end
 
-Base.size(A::AbstractArray2) = map(Int, ArrayInterface.size(A))
-Base.size(A::AbstractArray2, dim) = Int(ArrayInterface.size(A, dim))
+Base.size(A::AbstractArray2) = Base.size(Size(A))
+Base.size(A::AbstractArray2, dim) = length(Size(A, dim))
 
 function Base.axes(A::AbstractArray2)
     !(parent_type(A) <: typeof(A)) && return ArrayInterface.axes(parent(A))
@@ -731,13 +731,13 @@ function __init__()
         @generated function axes_types(::Type{<:StaticArrays.StaticArray{S}}) where {S}
             return Tuple{[StaticArrays.SOneTo{s} for s in S.parameters]...}
         end
-        @generated function size(A::StaticArrays.StaticArray{S}) where {S}
+        @generated function ArrayInterface.Size(A::StaticArrays.StaticArray{S}) where {S}
             t = Expr(:tuple)
             Sp = S.parameters
             for n = 1:length(Sp)
                 push!(t.args, Expr(:call, Expr(:curly, :StaticInt, Sp[n])))
             end
-            return t
+            return :(ArrayInterface.Size($t))
         end
         @generated function strides(A::StaticArrays.StaticArray{S}) where {S}
             t = Expr(:tuple, Expr(:call, Expr(:curly, :StaticInt, 1)))
--- a/src/size.jl
+++ b/src/size.jl
@@ -1,79 +1,115 @@
 
 """
-    size(A) -> Tuple
-    size(A, dim) -> Union{Int,StaticInt}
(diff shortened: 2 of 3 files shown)


Tokazama:
@ChrisRackauckas, this is related to the issue you raised here https://github.com/JuliaLang/julia/issues/25107

ChrisRackauckas:
No it isn't, because `Size` holds a tuple.

Tokazama:
`Size` would be equivalent to a dense constructor, so you could have something like:
```julia
construct(::CPUPointer, ::Type{T}, param::Size) where {T} = Array{T}(undef, size(param))
function construct(d::AbstractDevice, ::Type{T}, param::Banded) where {T}
    BandedMatrix(construct(d, T, Size(param)), bandwidths(param))
end

```

ChrisRackauckas:
Why not go straight to making a `AbstractConstructor` system?

Tokazama:
We could have `Size <: AbstractConstructor` and I'd actually prefer to have that, but due to [concerns about compilation latency](https://github.com/JuliaArrays/ArrayInterface.jl/issues/211) we've avoided added a system that would involve naturally involve too many additional types. I figured this approach would be a step in that direction without having to add on a whole new system all at once.

chriselrod:
JuliaSIMD already doesn't support ArrayInterface 4, but with this change it's increasingly unlikely that I'll find the time to add compatibility.

chriselrod:
I'll have to look at this more closely, but would packages defining `ArrayInterface.size` methods still be handled correctly?

ChrisRackauckas:
How are the downstream tests not failling?

Tokazama:
> I'll have to look at this more closely, but would packages defining `ArrayInterface.size` methods still be handled correctly?

That's the intention. If this just breaks a bunch of stuff then I don't think it would be worth implementing something like this until we have a more complete story for the whole construction interface.

chriselrod:
> How are the downstream tests not failling?

The tests pass if the compat of the downstream package excludes this ArrayInterface version.
At least in the case of LoopVectorization.jl, that is true (but sometimes Pkg thinks it is okay anyway?).

chriselrod:
> That's the intention. If this just breaks a bunch of stuff then I don't think it would be worth implementing something like this until we have a more complete story for the whole construction interface.

The tricky bit is that we'd need both `ArrayInterface.size` and `ArrayInterface.Size` to work correctly, regardless of which one someone implemented. This is of course circular:
```julia
Size(x) = Size(size(x))
size(x) = Size(x).size
```
One solution would be to use [static_hasmethod](https://github.com/oxinabox/Tricks.jl#we-can-use-static_hasmethod-to-declare-traits) to break the circle.

Tokazama:
Without an explicit method `Size` uses the length of `ArrayInterface.axes` which in turn falls back to `Base.axes`, so it should only be circular if `Base.axes(::ArrayType)` depends on `ArrayInterface.size(::ArrayType)`.

chriselrod:
> Without an explicit method, `Size` uses the length of `ArrayInterface.axes` which in turn falls back to `Base.axes`, so it should only be circular if `Base.axes(::ArrayType)` depends on `ArrayInterface.size(::ArrayType)`.

Okay, that sounds fine.

ChrisRackauckas:
@Tokazama could you please fix downstream. The way that recent changes has broken downstream makes it so that things cannot update and it is blocking the actual purpose of this package. If it doesn't get fixed soon I think we should just see adding some of the static stuff as a mistake and revert it / remove it all. It's supposed to be an interface for missing verbs, not something that is breaking constantly.

PR #242, opened 2022-02-11
Add issparsematrix

ChrisRackauckas:
Looks like this is a missing verb in the sparse matrix vocabulary which is required for writing generic sparse functionality that also works on the GPU. Here's 3 PRs which currently only work on the CPU because of the missing verb:

- https://github.com/JuliaDiff/SparseDiffTools.jl/pull/179
- https://github.com/SciML/LinearSolve.jl/pull/104
- https://github.com/JuliaDiff/FiniteDiff.jl/pull/130

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -22,6 +22,7 @@ ArrayInterface.is_lazy_conjugate
 ArrayInterface.ismutable
 ArrayInterface.issingular
 ArrayInterface.isstructured
+ArrayInterface.issparsematrix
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
@Tokazama I want to be able to use this downstream ASAP, so please fix your mess or I will need to revert it.

ChrisRackauckas:
But I still don't want master broken downstream for long. That's just a very bad sign.

Tokazama:
What mess are you referring to?

PR #243, opened 2022-02-14
Generalize from `Array` to `DenseArray` where possible in stridelayout

brenhinkeller:
This would help enable LoopVectorization of custom array types that subtype `<: DenseArray <: AbstractArray`. Since `DenseArray` is [defined to be column-major](https://docs.julialang.org/en/v1/manual/arrays/)

> DenseArray is an abstract subtype of AbstractArray intended to include all arrays where elements are stored contiguously in column-major order

this should be kosher AFAIU.


ChrisRackauckas:
https://github.com/JuliaArrays/ArrayInterface.jl/runs/5182799084?check_suite_focus=true#step:6:213
https://github.com/JuliaArrays/ArrayInterface.jl/runs/5182799146?check_suite_focus=true#step:6:172

@Tokazama We should revert on Friday if this all doesn't get fixed. Downstream shouldn't be failing for more than a week or two without a downstream PR.

src/stridelayout.jl
```diff
@@ -220,7 +220,7 @@ function stride_rank(::Type{T}) where {T}
         return stride_rank(parent_type(T))
     end
 end
-stride_rank(::Type{Array{T,N}}) where {T,N} = nstatic(Val(N))
+stride_rank(::Type{DenseArray{T,N}}) where {T,N} = nstatic(Val(N))
```
Tokazama, line 223: This needs to be subtyping `DenseArray`

`stride_rank(::Type{<:DenseArray{T,N}}) where {T,N} = nstatic(Val(N))`

brenhinkeller: Ah, that makes sense

brenhinkeller:
No rush on my end! This'll ultimately hopefully support LV of `MallocArray`s from StaticTools.jl but I've got lots of other things to worry about in the meanwhile, so can wait and see how https://github.com/JuliaSIMD/LoopVectorization.jl/pull/383 / the downstream fixes play out first

brenhinkeller:
Any progress on the downstream stuff lately @Tokazama? I'd try to help, but a lot of the LV internals are still beyond me

Tokazama:
Sorry I've been recovering from COVID the last week and haven't been very active on GitHub. LoopVectorization has some odd failures that I don't understand why they occur with the change in question.

Tokazama:
This is an inference issue we've ran into on nightly

chriselrod (approved):

Tokazama:
@brenhinkeller, if you think this should be released in a new update then there needs to be an update to Project.toml too.

brenhinkeller:
Version updated!

PR #244, opened 2022-02-23
Fix spelling

YingboMa:


YingboMa:
The rest of the file use `isstructured` https://github.com/JuliaArrays/ArrayInterface.jl/blob/75c9a32b7b55de977cab1a847a462ee0665999dd/src/ArrayInterface.jl#L204-L218

ChrisRackauckas (approved):

PR #251, opened 2022-03-13
Add `can_setindex` methods

Tokazama:


chriselrod (approved):
Of course, some of these can `Base.setindex`, but `can_setindex` is already about `setindex!` instead.

src/ArrayInterface.jl
```diff
@@ -187,6 +187,11 @@ Query whether a type can use `setindex!`.
 can_setindex(x) = can_setindex(typeof(x))
 can_setindex(::Type) = true
 can_setindex(::Type{<:AbstractRange}) = false
+can_setindex(::Type{<:AbstractDict}) = true
+can_setindex(::Type{<:Base.ImmutableDict}) = false
+can_setindex(@nospecialize T::Type{<:Tuple}) = false
+can_setindex(@nospecialize T::Type{<:NamedTuple}) = false
+can_setindex(::Type{<:Base.Pairs{<:Any,<:Any,P}}) where {P} = can_setindex(P)
```
chriselrod, line 194: Not defined in Julia 1.6, apparently.

Tokazama:
> Of course, some of these can `Base.setindex`, but `can_setindex` is already about `setindex!` instead.

Is your point that `can_setindex` could be confused as referring to `Base.setindex`?

src/ArrayInterface.jl
Tokazama: Just realized it's in `Iterators.Pairs`

chriselrod:
> > Of course, some of these can `Base.setindex`, but `can_setindex` is already about `setindex!` instead.
> 
> Is your point that `can_setindex` could be confused as referring to `Base.setindex`?

Yeah. We already have `can_setindex` meaning this though. Was just making an observation.

Tokazama:
> > > Of course, some of these can `Base.setindex`, but `can_setindex` is already about `setindex!` instead.
> > 
> > 
> > Is your point that `can_setindex` could be confused as referring to `Base.setindex`?
> 
> Yeah. We already have `can_setindex` meaning this though. Was just making an observation.

It's a good point and looking at the blame history I don't think it was discussed in much detail when implemented. I can make an issue for it just in case anyone feels strongly about this or wants to approach that issue in another PR.

ChrisRackauckas:
And one of the most important cases for this would be CuArrays, which should get an overload that is dependent on the global.

Tokazama:
I'm assuming that would have to be implemented with something like `CUDA.context`.

PR #259, opened 2022-04-01
Rename static_length to length and document

### LINKED ISSUE #254: `ArrayInterface.length`?
I decided to try out `ArrayInterface` for a package I am writing and I have found it very useful.

I was surprised to find that there is no `ArrayInterface.length` that is analogous to `ArrayInterface.size`.  For example,
```julia
julia> using ArrayInterface, StaticArrays

julia> A = SMatrix{2,2}(1:4)
2×2 SMatrix{2, 2, Int64, 4} with indices SOneTo(2)×SOneTo(2):
 1  3
 2  4

julia> ArrayInterface.size(A)
(static(2), static(2))

julia> ArrayInterface.length(A)
4
```
I expected last result to be `static(4)` rather than `4` since all the dimensions of `A` are static.  However, it turns out that `ArrayInterface.length === Base.length`.  Why not define something like `ArrayInterface.length(A) = reduce(*, ArrayInterface.size(A))`, which would give `static(4)` in the example above?

One might also have cases where the length of an array is known but the size of one or more of the dimensions is not.  For example, I think the output of `reshape(A, ...)` would usually be like this because the `length` is invariant but the `size` is not.  In fact,
```julia
ArrayInterface.known_length(ArrayInterface.size(reshape(A, :)))
```
returns `nothing` even though it really is known.  Something more sophisticated than my suggestion above would be required to handle this case though.

nchisholm:
I wasn't 100% sure how to best handle places where there was an `===` comparison to the result of `Base.length(x)`.  I ended up changing these to `dynamic(ArrrayInterface.length(x)) === ...`, but I'm not sure that is the preferred way.  Happy to make adjustments.

Also I'm not very familiar with Documenter and was getting the following warning:

```
┌ Warning: 3 docstrings not included in the manual:
│ 
│     ArrayInterface.SUnitRange
│     ArrayInterface.SOneTo
│     ArrayInterface.length :: Union{Tuple{UnitRange{T}}, Tuple{T}} where T
│ 
│ 
│ These are docstrings in the checked modules (configured with the modules keyword)
│ that are not included in @docs or @autodocs blocks.
```

Resolves #254

### DIFF
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -36,8 +36,26 @@ const MatAdjTrans{T,M<:AbstractMatrix{T}} = Union{Transpose{T,M},Adjoint{T,M}}
 const UpTri{T,M} = Union{UpperTriangular{T,M},UnitUpperTriangular{T,M}}
 const LoTri{T,M} = Union{LowerTriangular{T,M},UnitLowerTriangular{T,M}}
 
-@inline static_length(a::UnitRange{T}) where {T} = last(a) - first(a) + oneunit(T)
-@inline static_length(x) = Static.maybe_static(known_length, length, x)
+"""
+    length(A) -> Union{Int,StaticInt}
+
+Returns the length of `A`.  If the length is known at compile time, it is
+returned as `Static` number.  Otherwise, `ArrayInterface.length(A)` is identical
+to `Base.length(A)`.
+
+```julia
+julia> using StaticArrays, ArrayInterface
+
+julia> A = @SMatrix rand(3,4);
+
+julia> ArrayInterface.length(A)
+static(12)
+```
+"""
+@inline length(a::UnitRange{T}) where {T} = last(a) - first(a) + oneunit(T)
+@inline length(x) = Static.maybe_static(known_length, Base.length, x)
+const static_length = length  # for backward compatibility
+
 @inline static_first(x) = Static.maybe_static(known_first, first, x)
 @inline static_last(x) = Static.maybe_static(known_last, last, x)
 @inline static_step(x) = Static.maybe_static(known_step, step, x)
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -54,7 +54,9 @@ static(12)
 """
 @inline length(a::UnitRange{T}) where {T} = last(a) - first(a) + oneunit(T)
 @inline length(x) = Static.maybe_static(known_length, Base.length, x)
-const static_length = length  # for backward compatibility
+
+# Alias to to-be-depreciated internal function
+const static_length = length
 
 @inline static_first(x) = Static.maybe_static(known_first, first, x)
 @inline static_last(x) = Static.maybe_static(known_last, last, x)
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -243,7 +243,7 @@ indices calling [`to_axis`](@ref).
 @inline function to_axes(A, inds::Tuple)
     if ndims(A) === 1
         return (to_axis(axes(A, 1), first(inds)),)
-    elseif dynamic(length(inds)) === 1
+    elseif Base.length(inds) === 1
         return (to_axis(eachindex(IndexLinear(), A), first(inds)),)
     else
         return to_axes(A, axes(A), inds)
@@ -378,14 +378,14 @@ end
 _ints2range(x::Integer) = x:x
 _ints2range(x::AbstractRange) = x
 @inline function unsafe_get_collection(A::CartesianIndices{N}, inds) where {N}
-    if (dynamic(length(inds)) === 1 && N > 1) || stride_preserving_index(typeof(inds)) === False()
+    if (Base.length(inds) === 1 && N > 1) || stride_preserving_index(typeof(inds)) === False()
         return Base._getindex(IndexStyle(A), A, inds...)
     else
         return CartesianIndices(to_axes(A, _ints2range.(inds)))
     end
 end
 @inline function unsafe_get_collection(A::LinearIndices{N}, inds) where {N}
-    if dynamic(length(inds)) === 1 && isone(_ndims_index(typeof(inds), static(1)))
+    if Base.length(inds) === 1 && isone(_ndims_index(typeof(inds), static(1)))
         return @inbounds(eachindex(A)[first(inds)])
     elseif stride_preserving_index(typeof(inds)) === True()
         return LinearIndices(to_axes(A, _ints2range.(inds)))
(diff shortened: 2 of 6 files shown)


src/indexing.jl
```diff
@@ -243,7 +243,7 @@ indices calling [`to_axis`](@ref).
 @inline function to_axes(A, inds::Tuple)
     if ndims(A) === 1
         return (to_axis(axes(A, 1), first(inds)),)
-    elseif length(inds) === 1
+    elseif dynamic(length(inds)) === 1
```
Tokazama, line 246: You can just do `Base.length(inds)` here to avoid having to explicitly convert to `Int`.

src/indexing.jl
```diff
@@ -378,14 +378,14 @@ end
 _ints2range(x::Integer) = x:x
 _ints2range(x::AbstractRange) = x
 @inline function unsafe_get_collection(A::CartesianIndices{N}, inds) where {N}
-    if (length(inds) === 1 && N > 1) || stride_preserving_index(typeof(inds)) === False()
+    if (dynamic(length(inds)) === 1 && N > 1) || stride_preserving_index(typeof(inds)) === False()
```
Tokazama, line 381: Same here.

src/indexing.jl
```diff
@@ -378,14 +378,14 @@ end
...
         return Base._getindex(IndexStyle(A), A, inds...)
     else
         return CartesianIndices(to_axes(A, _ints2range.(inds)))
     end
 end
 @inline function unsafe_get_collection(A::LinearIndices{N}, inds) where {N}
-    if length(inds) === 1 && isone(_ndims_index(typeof(inds), static(1)))
+    if dynamic(length(inds)) === 1 && isone(_ndims_index(typeof(inds), static(1)))
```
Tokazama, line 388: Same here

src/ArrayInterface.jl
```diff
@@ -36,8 +36,26 @@ const MatAdjTrans{T,M<:AbstractMatrix{T}} = Union{Transpose{T,M},Adjoint{T,M}}
 const UpTri{T,M} = Union{UpperTriangular{T,M},UnitUpperTriangular{T,M}}
 const LoTri{T,M} = Union{LowerTriangular{T,M},UnitLowerTriangular{T,M}}
 
-@inline static_length(a::UnitRange{T}) where {T} = last(a) - first(a) + oneunit(T)
-@inline static_length(x) = Static.maybe_static(known_length, length, x)
```
Tokazama, line 40: Technically this isn't part of the public API, but just in case it might be good to do `@deprecate static_length(x) = ArrayInterface.length(x)`.

Tokazama (changes requested):

src/ArrayInterface.jl
chriselrod: The JuliaSIMD ecosystem has been using `ArrayInterface.static_length`, e.g.
https://github.com/JuliaSIMD/LoopVectorization.jl/search?q=static_length
so I would prefer a major version bump and PRs to these packages to update them.

ChrisRackauckas:
We should schedule the breaking changes and not be so quick with them. We just had v5 not so long ago 😅

nchisholm:
Thanks for your quick feedback/review.  Would it be an option to wait on depreciating the old name but add the new part of the interface?  I did alias `const static_length = length` to try to avoid unexpected breakage, but I understand if it is a better option to wait on this. Let me know what I should do.

Also, this is my first PR, so thanks for your patience!  I am still figuring things out a bit. (I regret a conflict with @ChrisRackauckas's recent hackathon on this - would have been useful!).

src/ArrayInterface.jl
Tokazama: Yeah, that's what I was afraid of

Tokazama:
> I did alias `const static_length = length` to try to avoid unexpected breakage, but I understand if it is a better option to wait on this.

Sorry, I didn't catch the constant definition. The whole deprecation thing may not matter then.

src/ArrayInterface.jl
```diff
@@ -36,8 +36,26 @@ const MatAdjTrans{T,M<:AbstractMatrix{T}} = Union{Transpose{T,M},Adjoint{T,M}}
...
+
+julia> A = @SMatrix rand(3,4);
+
+julia> ArrayInterface.length(A)
+static(12)
+```
+"""
+@inline length(a::UnitRange{T}) where {T} = last(a) - first(a) + oneunit(T)
+@inline length(x) = Static.maybe_static(known_length, Base.length, x)
+const static_length = length  # for backward compatibility
```
nchisholm, line 57: Should this sufficiently protect against breakage of dependent packages?

chriselrod: Yes.

ChrisRackauckas:
I guess for now, const it and open an issue saying that we should do the deprecation in the future when we consider v6.

Tokazama:
@nchisholm, my request to change `dynamic(length(x))` to `Base.length(x)` is mostly because I think it makes it easier to read for people not familiar with `Static.jl`, but if others disagree I'm not going to put up any fight in this getting merged. I really appreciate your enthusiasm here.

nchisholm:
> my request to change dynamic(length(x)) to Base.length(x) is mostly because I think it makes it easier to read for people not familiar with Static.jl, but if others disagree I'm not going to put up any fight in this getting merged.

I wasn't sure to what degree `ArrayInterface` should be independent of the interface in `Base` if the eventual goal is to replace it, but I imagine that's pretty far off.

> I really appreciate your enthusiasm here.

Thanks! Happy to contribute what I can.

Tokazama:
There's a bit of stuff in limbo around that right now. It sounds like we'll have a better idea of which things could move to base when Julia 1.9 comes around.

src/ArrayInterface.jl
```diff
@@ -36,8 +36,26 @@ const MatAdjTrans{T,M<:AbstractMatrix{T}} = Union{Transpose{T,M},Adjoint{T,M}}
...
+```julia
+julia> using StaticArrays, ArrayInterface
+
+julia> A = @SMatrix rand(3,4);
+
+julia> ArrayInterface.length(A)
+static(12)
+```
+"""
+@inline length(a::UnitRange{T}) where {T} = last(a) - first(a) + oneunit(T)
```
nchisholm, line 55: One last question before I push a revision. I am wondering why there is a specific method for `UnitRange` here.  I tried removing it and everything seemed to be fine.

Tokazama: @chriselrod added that a while back when the underlying call in base was overly complicated and slow. I'm pretty sure that has changed since. However, `UnitRange` can't ever have a static length so I'd leave that to Chris to change if it's no longer necessary.

Tokazama (commented):
LGTM

Tokazama (approved):

Tokazama:
@chriselrod, Are you fine with this?

chriselrod:
> @chriselrod, Are you fine with this?

Yes.

chriselrod (approved):

Issue #211, opened 2021-10-07
ArrayInterfaceCore


ParadaCarleton:
As mentioned [here](https://github.com/rafaqz/DimensionalData.jl/issues/307), `ArrayInterface` has become a huge dependency to take on. It might make sense to separate out some `ArrayInterfaceCore` that can be loaded faster and contains only basic functionality.

grahamas:
Sounds like:

1. `AbstractArrayInterface` [1]
2. and `ArrayInterfaceCore` 

I think there's a more useful divide. It seems to me like there are two kinds of users:

1. Someone purely trying to develop their own Array type (usually the ideal use-case for `AbstractX` packages). However, even for this person the `AbstractArrayInterface` wouldn't be sufficient in many cases. The existing package provides a lot of useful low-level stuff (`device` etc) that a more advanced developer might want to use without the cost of importing the various high-level dependencies of this package. This is the person who probably wants something like `ArrayInterfaceCore`, for some definition of "basic functionality," and so is the subject of this issue.
2. Someone trying to use one (or more) of the Array types defined in this package. This person wants more than would be included in my understanding of "basic functionality" (maybe e.g. the `FieldArray`) but still a lot less than the full package (e.g. does not want `CUDA`, `CuArray` etc.). This person suggest splitting up ArrayInterface into sub-packages, all depending on `ArrayInterfaceCore`, each contributing one set of Array implementations. This person isn't really relevant to this issue, but sometimes this person is me, so I think it'd be cool. But it should be clear/decided that "basic functionality" won't include this person.

Basically, that's all to say I don't think `AbstractArrayInterface` (the package) would be all that useful, and I think "basic functionality" can mean a lot of different things to different people. From my perspective, I'd like it to mean that when I subtype `AbstractArrayInterface`, the vast majority of necessary and optional methods have working default implementations. Someone should probably write down what it means, though.

Finally, I'm curious how this ties into the long-mentioned merge into base julia.

[1]  the package, not the type--- maybe this is why no one suggested this.

ParadaCarleton:
> What's the measure here? Load time?

I think it should be; anything like an ArrayInterfaceCore should take under 0.1s to load, I think. Is there a breakdown of load time by file? What precisely makes ArrayInterface take so long to load?

ChrisRackauckas:
Julia v1.7?

ChrisRackauckas:
[...]

PR #263, opened 2022-05-09
Multi package (#211)

Tokazama:


src/ArrayInterface.jl
```diff
@@ -1,735 +1,18 @@
...
-                Ptr{Cvoid}(),
-                Ptr{Cvoid}(),
-                1,
-                1,
-                jac_prototype.colptr[1:1],
-                jac_prototype.rowval[1:1],
-                jac_prototype.nzval[1:1],
-                0,
-            )
-        end
```
ChrisRackauckas, line 730: Where did this one go?

Tokazama: It's part of the standard library so I put it in ArrayInterfaceCore

chriselrod:
My vision of the "core" was that it'd be stubs other packages can depend on to provide methods without sacrificing load time when ArrayInterface wasn't used.

At the very least, `Requires` should be avoided for light weight libraries.

src/ArrayInterface.jl
chriselrod: Probably better to just depend on SuiteSparse than depend on Requires.

chriselrod: A quick look at the code/Project.toml shows you're depending on SparseArrays but not Requires.

Tokazama:
> My vision of the "core" was that it'd be stubs other packages can depend on to provide methods without sacrificing load time when ArrayInterface wasn't used.
> 
> At the very least, `Requires` should be avoided for light weight libraries.

ArrayInterfaceCore doesn't depend on Requires at all, which drastically improves load time. We could have `ArrayInterfaceCore`, `ArrayInterfaceBase`, and `ArrayInterface` so that we can have just provide stubs. I'm not sure it's very useful to provide methods that don't support base types at all though.

johnnychen94:
Curiously, `using ArrayInterface` time drops from 100ms to 60ms, is this also expected?

ChrisRackauckas:
Yes, the static stuff isn't very core, it's not just the stub functions. The core should probably have only standard library dependencies. Things like:

https://github.com/JuliaArrays/ArrayInterface.jl/blob/master/src/ArrayInterface.jl#L204-L216

should be what the core looks like, just simple free definitions of interface functions to be extended. Then ArrayInterfaceStatic would add all of the heavier static stuff (well, at least until StaticInt doesn't invalidate). ArrayInterfaceCUDA, ArrayInterfaceBandedMatrices, etc.

Tokazama:
So instead of using requires at all have individual packages for each requires section?

ChrisRackauckas:
Yes, the main thing should be that Requires.jl usage should be completely eliminated.

johnnychen94:
Since there's a trend to split out many interface packages here, I'd also like to ask how you guys think of https://discourse.julialang.org/t/using-the-type-and-functions-for-dispatch-purposes-without-loading-all-codes-of-its-owner-package/79951 question. Should this be a missing tooling issue, or should this be supported somewhere at language or Pkg level?

ChrisRackauckas:
Multi-packages in a single repo is a nice way to handle this development-wise, so I think it's fine.

Tokazama:
> Since there's a trend to split out many interface packages here, I'd also like to ask how you guys think of https://discourse.julialang.org/t/using-the-type-and-functions-for-dispatch-purposes-without-loading-all-codes-of-its-owner-package/79951 question. Should this be a missing tooling issue, or should this be supported somewhere at language or Pkg level?

I think a core dev proposed a sort of header file for Julia recently in a gist, but organizing and developing interfaces probably doesn't have a one size fits all solution.

Tokazama:
> Then ArrayInterfaceStatic would add all of the heavier static stuff (well, at least until StaticInt doesn't invalidate). ArrayInterfaceCUDA, ArrayInterfaceBandedMatrices, etc.

`StaticInt` is pretty well integrated into the package, so I'm not sure how to make a meaningful division between `ArrayInterfaceStatic` and `ArrayInterfaceCore`.

ChrisRackauckas:
> StaticInt is pretty well integrated into the package

But it's not. The vast majority of packages don't use those functions directly, instead it's a smaller set (mostly JuliaSIMD) that uses the static functions. So either fix the invalidations (preferred) or section it off.

https://github.com/JuliaArrays/ArrayInterface.jl/issues/121#issuecomment-779910933

I think it's time to make this change.

Tokazama:
This https://github.com/JuliaLang/julia/pull/44538 would fix the invalidations.

These can probably be separated from anything in Static.jl

* allowed_getindex
* allowed_setindex!
* parent_type
* ismutable
* can_change_size
* restructure
* can_avx
* known_last
* known_first
* known_step
* can_setindex
* aos_to_soa
* zeromatrix
* lu_instance
* fast_matrix_colors
* findstructralnz
* BidiagonalIndex
* TridiagonalIndex
* ArrayIndex
* issingular
* fast_scalar_indexing
* has_sparsestruct
* isstructured
* safevec

These use Static.jl either directly or indirectly in some way.

* contiguous_axis
* contiguous_axis_indicator
* contiguous_batch_size
* defines_strides
* device
* dimnames
* has_dimnames
* has_parent
* is_column_major
* is_lazy_conjugate
* is_splat_index
* known_dimnames
* known_length
* known_offsets
* known_offset1
* known_size
* known_strides
* ndims_index
* axes
* axes_types
* broadcast_axis
* deleteat
* dense_dims
* from_parent_dims
* getindex
* indices
* insert
* lazy_axes
* offset1
* offsets
* setindex!
* size
* strides
* to_axes
* to_axis
* to_dims
* to_index
* to_indices
* to_parent_dims
* unsafe_reconstruct
* BroadcastAxis
* LazyAxis
* OptionallyStaticStepRange
* OptionallyStaticUnitRange
* SOneTo
* SUnitRange
* StrideIndex

I'm working on a way to section off the latter in a way that's more meaningful than just arbitrarily throwing anything the happens to use Static.jl, but ultimately it would be best to have the aforementioned PR to Base merged.

libs/ArrayInterfaceCuArrays/Project.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[deps]
+Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
+ArrayInterfaceCore = "30b0a656-2188-435a-8636-2ec0e6a096e2"
+CuArrays = "3a865a2d-5b23-5a0f-bc46-62713ec82fae"
+
+[compat]
+Adapt = "3"
+ArrayInterfaceCore = "0.1"
+CuArrays = "2.2"
```
ToucheSir, line 13: CuArrays being long deprecated in favour of CUDA, could this whole package be pruned?

Tokazama: We only support the latest LTS of Julia so we may be inadvertently not supporting it as is

ChrisRackauckas: Yeah just delete CuArrays stuff.

ChrisRackauckas:
I setup all of the tests and am giving this a go, once tests pass I'll merge. I still want to separate out the static stuff from core, but I think this is a good merge point instead of trying to stuff more in here since it's a huge change.

chriselrod:
So, the problem is that ArrayInterfaceCore is not registered, and therefore integration tests can't run?

Tokazama:
Thanks Chris Rackauckas for doing this. I've looked at load times and have seen a lot of improvement, but `ArrayInterfaceStaticArrays` seems to remain an issue. I'll try improving that later on as we get everything else working.

chriselrod:
VectorizationBase test failures:
```julia
LazyMulAdd Loads/Stores: Error During Test at /home/chriselrod/.julia/dev/VectorizationBase/test/runtests.jl:542
  Got exception outside of a @test
  "Memory access for OffsetArray{Float64, 3, PermutedDimsArray{Float64, 3, (3, 1, 2), (2, 3, 1), Array{Float64, 3}}} not implemented yet."
  Stacktrace:
    [1] memory_reference
      @ ~/.julia/packages/LayoutPointers/oZEWc/src/stridedpointers.jl:45 [inlined]
```

Tokazama:
> I'm trying to run tests locally. LoopVectorization fails to precompile:
> 
> ```julia
> Precompiling project...
>   ✗ LoopVectorization
>   0 dependencies successfully precompiled in 2 seconds (57 already precompiled)
>   1 dependency errored. To see a full report either run `import Pkg; Pkg.precompile()` or load the package
>      Testing Running tests...
> ERROR: LoadError: UndefVarError: OptionallyStaticRange not defined
> Stacktrace:
> ```
> 
> So it and everything depending on it fail.
> 
> https://github.com/JuliaSIMD/LoopVectorization.jl/blob/1fdc689ec40132c10014b564269101be92e5280f/src/LoopVectorization.jl#L131-L133
> 
> Also, `UpTri` is not defined.

I'll go through those imports tonight. It would be good if we were more strict about what is considered public API but now's not the time to fight that battle

src/ArrayInterface.jl
```diff
@@ -31,6 +31,12 @@ import ArrayInterfaceCore: ismutable, can_change_size, can_setindex, deleteat, i
 # constants
 import ArrayInterfaceCore: MatAdjTrans, VecAdjTrans, UpTri, LoTri
 
+using ArrayInterfaceBandedMatrices
+using ArrayInterfaceBlockBandedMatrices
+using ArrayInterfaceCUDA
+using ArrayInterfaceOffsetArrays
+using ArrayInterfaceStaticArrays
+using ArrayInterfaceTracker
```
ChrisRackauckas, line 39: No, those were removed on purpose. It would be really breaking to the ecosystem if ArrayInterface just started requiring CUDA and all of that. Instead, ArrayInterface should be what we're calling ArrayInterfaceStatic, allowing the core to just be the minimal definitions and quick methods that have no loading impact (i.e., all of the non-static stuff).

ChrisRackauckas:
So here's the deal. Registering packages for the first time is a 3 day process. Let's get that going so that ArrayInterfaceCore is going, can do its 3 days, and then most of the subpackages can go.

The plan is to get ArrayInterfaceCore to be an essentially instant startup, and anything with the static stuff can go into ArrayInterface. A version of ArrayInterface with all of the dependencies I don't think makes sense for any application, so a "somewhat bigger ArrayInterface" is a good use for the higher package, leaving Core with just the interface definitions and some small functions that don't have a load time impact (and 0 dependencies beyond the standard library).

So let's get this started. The next ArrayInterface will be given a major bump anyways (since it's breaking to remove those Requires.jl support parts and now require subpackages to be used), and registering ArrayInterfaceCore won't be breaking to anyone (since it doesn't exist right now), so this gives a safe path. And we should not do the ArrayInterface major bump until downstream all looks good (mainly, JuliaSIMD downstream since that's what uses the static stuff. All of SciML should be fine moving to Core).

chriselrod:
> So here's the deal. Registering packages for the first time is a 3 day process. Let's get that going so that ArrayInterfaceCore is going, can do its 3 days, and then most of the subpackages can go.
> 
> The plan is to get ArrayInterfaceCore to be an essentially instant startup, and anything with the static stuff can go into ArrayInterface. A version of ArrayInterface with all of the dependencies I don't think makes sense for any application, so a "somewhat bigger ArrayInterface" is a good use for the higher package, leaving Core with just the interface definitions and some small functions that don't have a load time impact (and 0 dependencies beyond the standard library).
> 
> So let's get this started. The next ArrayInterface will be given a major bump anyways (since it's breaking to remove those Requires.jl support parts and now require subpackages to be used), and registering ArrayInterfaceCore won't be breaking to anyone (since it doesn't exist right now), so this gives a safe path. And we should not do the ArrayInterface major bump until downstream all looks good (mainly, JuliaSIMD downstream since that's what uses the static stuff. All of SciML should be fine moving to Core).

We can of course have `ArrayInterfaceCore.size` without `static`, but also methods like `ArrayInterfaceCore.contiguous_axis`:
```julia
module ArrayInterfaceCore
"""
    $(helpful_documentation)
"""
function contiguous_axis end

end # module

module ArrayInterface

using ArrayInterfaceCore, Static

# pirate
ArrayInterfaceCore.contiguous_axis(::Type{<:Array}) = static(1)
end # module
```
so people who just want to provide the interface but not use it (e.g., all they want is LoopVectorization to work on their array types) can get by with only `ArrayInterfaceCore` and `Static`.

ChrisRackauckas:
Definitely blank method definitions can be in the core. Yes, I think every interface definition should be there. The only things that are not blank functions should be "trivial" functionality, I think this is a good example:

https://github.com/JuliaArrays/ArrayInterface.jl/blob/master/lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl#L52-L62

Having those dispatches in the Core will not change compilation or pretty much anything at all, so it's fine to give that a 👍 for Core and not require the full for this. Things like:

https://github.com/JuliaArrays/ArrayInterface.jl/blob/master/lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl#L146-L158

should stay too then, since it's a nice concise interface definition and application in the standard library. 

What I think shouldn't be core are the implementations of some of those static callbacks which take up entire files of their own. That's what then takes up load time, and that should be "okay, grab the rest". It's a judgement call where that line is, but the piracy of Core approach is definitely what we need to do there (and that's why it should be in the same repo).

Tokazama:
When I tested load times for ArrayInterfaceCore it looked like it was ~.01 seconds on top of Static.jl. I'm not sure if I have strong opinions on how this is improved further right now, but it might be worth actually experimenting with this more before committing to an entirely new direction. I have some loose ideas on how to improve things further, but I'm finishing up a crazy clinic schedule right now so I haven't put together any empirical evidence that one approach will work.

ChrisRackauckas:
Yeah so maybe all that needs to be done is to move all of the functions that require Static.jl to ArrayInterface proper.

PR #264, opened 2022-05-16
Fix `strides(::ReinterpretArray)` and extend `strides(::ReshapedArray)`

N5N3:
Previous implement is somehow limited (and wrong in some cases).
This PR tries to fix and extend them.

### DIFF
--- a/src/stridelayout.jl
+++ b/src/stridelayout.jl
@@ -262,31 +262,29 @@ stride_rank(x, i) = stride_rank(x)[i]
 function stride_rank(::Type{R}) where {T,N,S,A<:Array{S},R<:Base.ReinterpretArray{T,N,S,A}}
     return nstatic(Val(N))
 end
-if VERSION ≥ v"1.6.0-DEV.1581"
-  @inline function stride_rank(::Type{A}) where {NB, NA, B <: AbstractArray{<:Any,NB},A<: Base.ReinterpretArray{<:Any, NA, <:Any, B, true}}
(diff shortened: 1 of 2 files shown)


Tokazama:
Thanks for the PR! It looks like this is changing some of the tests so we'll probably want to wait on reviewing this until #263 is merged since we're in the midst of a lot of big changes as is.

N5N3:
Looks like the `strides` api is not tested on master?

chriselrod:
> I'm not sure what the trade offs are to determine if something is strided at run time. 

Where is this being determined at runtime?
As a function of runtime values?

It'd have uses, but I'd worry about type instabilities of course.
It'd be something worth branching on in many cases.

At a glance, this PR LGTM.

N5N3:
> It'd have uses, but I'd worry about type instabilities of course.

I tried my best to ensure internal's stabilities. But the current version was modified based on `Base.strides`, which has no `StaticInt`. So I'm not sure the added test is surficent.

BTW, do we need to wait for https://github.com/SciML/Static.jl/pull/64?

Tokazama:
> > It'd have uses, but I'd worry about type instabilities of course.
> 
> I tried my best to ensure internal's stabilities. But the current version was modified based on `Base.strides`, which has no `StaticInt`. So I'm not sure the added test is surficent.

I guess the actual number of strides will always be known at compile time because it will always be the dimensions of the most array `strides` is called on and think everything goes to `Int` anyway. We do tend to find odd inference bugs when some downstream package has a deeply nested call to stuff, so we may need to call on you again if something pops up.

> BTW, do we need to wait for [SciML/Static.jl#64](https://github.com/SciML/Static.jl/pull/64)?

Just replace `Integer` with `Union{Int,StaticInt}`. That'll work now and later.

PR #273, opened 2022-05-22
Bump version to 6.0.1

chriselrod:
And release to hopefully resolve https://github.com/JuliaSIMD/VectorizationBase.jl/runs/6541055845?check_suite_focus=true ?
Maybe we need a new ArrayInterfaceCore release, too?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterface"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "6.0.0"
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
What would this bump?

chriselrod:
> What would this bump?

Things "just worked" locally when on master, so I assumed the released versions are out of date/broken.

It's hard to navigate/find out what the released versions are of a mono repo, so shooting randomly in the dark seems like the most straightforward strategy, other than of course splitting this up into a bunch of separate repos.

PR #275, opened 2022-05-24
Simplify implementation of known_length

Tokazama:


ChrisRackauckas:
It's static-free now?

ChrisRackauckas:
Is this downstream failure related?

https://github.com/SciML/SciMLBase.jl/runs/6575690017?check_suite_focus=true#step:6:135

Tokazama:
I don't think so, but I can look into it right now.

ChrisRackauckas:
That turned out to be just bad versioning.

PR #276, opened 2022-05-25
Move known_first, known_step, and known_last to ArrayInterfaceCore

Tokazama:
These don't require Static.jl at all to work

Edit:

I figured I'd move a couple other things over that I noticed didn't require Static.jl but I found some docstrings missing, so this is a draft until I've gone through everything once more.

BTW I'm ensuring we keep load time down down for ArrayInterfaceCore while this is happening (this PR has it at 0.006451 seconds on v1.7).


Tokazama:
I think failures may be related to https://github.com/JuliaArrays/ArrayInterface.jl/issues/268 because it's working

ChrisRackauckas:
Yeah I really need some help fixing up the CI here. I can test it locally as well.

Tokazama:
I'll take a look at it tonight but tbh I've never had much luck figuring out peculiarities with CI

Tokazama:
I fixed downstream testing for the most part. It appears that the errors in LoopVectorization are related to stale dependencies and compat bounds from aqua.

chriselrod:
> I fixed downstream testing for the most part. It appears that the errors in LoopVectorization are related to stale dependencies and compat bounds from aqua.
> 
> @chriselrod, could you just do a quick double check on the tests for LoopVectorizationBase to ensure I'm interpreting this correctly and not breaking anything?

Yeah, those seem to be the only failures.

Tokazama:
> Aqua seems to be confused/looking at the wrong Project.tomls in these errors?

It kind of looks like that but I'm still struggling to figure out why

Tokazama:
Besides OrdinaryDiffEq being stuck I think I it's all working now

.github/workflows/Downstream.yml
```diff
@@ -26,24 +26,13 @@ jobs:
           - {user: SciML, repo: OrdinaryDiffEq.jl, group: Core}
           - {user: SciML, repo: OrdinaryDiffEq.jl, group: Interface}
           - {user: SciML, repo: DelayDiffEq.jl, group: Interface}
-          - {user: JuliaSIMD, repo: VectorizationBase.jl, group: Interface}
-          - {user: JuliaSIMD, repo: LoopVectorization.jl, group: Downstream3}
```
ChrisRackauckas, line 30: why were these removed?

.github/workflows/Downstream.yml
```diff
@@ -56,7 +45,9 @@ jobs:
           try
             # force it to use this PR's version of the package
             Pkg.develop(PackageSpec(path="."))  # resolver may fail with main deps
-            
+            Pkg.develop(Pkg.PackageSpec(path=joinpath(pwd(), "lib", "ArrayInterfaceCore")))
+            Pkg.develop(Pkg.PackageSpec(path=joinpath(pwd(), "lib", "ArrayInterfaceOffsetArrays")))
+            Pkg.develop(Pkg.PackageSpec(path=joinpath(pwd(), "lib", "ArrayInterfaceStaticArrays")))
```
ChrisRackauckas, line 50: why were these two added?

.github/workflows/Downstream.yml
Tokazama: They were moved to LoopVecTests because Aqua was acting funny I'll see if I can consolidate it again.

ChrisRackauckas (approved):

Tokazama:
@chriselrod, LV was passing tests the last go around and I didn't change any actual code. I also just tested it with these changes locally and it worked so I'm assuming that error is unrelated this PR. Are you good with this being merged?

chriselrod:
> @chriselrod, LV was passing tests the last go around and I didn't change any actual code. I also just tested it with these changes locally and it worked so I'm assuming that error is unrelated this PR. Are you good with this being merged?

Yes, that test should be using `isapprox` anyway.

chriselrod (approved):

PR #281, opened 2022-05-31
Fix failing tests

Tokazama:
I think the problem we've had with testing the "lib/*" packages has been `Pkg.develop("ArrayInterface")` would pull from github in the ".ci/test.yml". Once I got that working I found a few more bugs from earlier PRs today, so I just went ahead and fixed those here too.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -40,7 +40,6 @@ for g in groups
         include("dimensions.jl")
         include("indexing.jl")
         include("ranges.jl")
-        include("setup.jl")
         include("size.jl")
         include("misc.jl")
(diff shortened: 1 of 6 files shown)


test/runtests.jl
```diff
@@ -40,7 +40,6 @@ for g in groups
         include("dimensions.jl")
         include("indexing.jl")
         include("ranges.jl")
-        include("setup.jl")
```
chriselrod, line 43: what was this file?

Tokazama: It's the first file included in that list. We were just including it twice.

PR #286, opened 2022-06-01
Pre static changes

Tokazama:
Motivated by https://github.com/SciML/Static.jl/pull/64.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterface"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "6.0.5"
+version = "6.0.6"
 
 [deps]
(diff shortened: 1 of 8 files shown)


chriselrod:
Did you see
```julia
WARNING: Method definition promote_shape(Tuple{Vararg{Union{Int64, Static.StaticInt{N} where N}}}, Tuple{Vararg{Union{Int64, Static.StaticInt{N} where N}}}) in module Static at /home/chriselrod/.julia/dev/Static/src/Static.jl:274 overwritten in module ArrayInterface at /home/chriselrod/.julia/dev/ArrayInterface/src/dimensions.jl:16.
  ** incremental compilation may be fatally broken for this module **
```
?

Tokazama:
That was the intention, since `promote_shape` doesn't require anything but `StaticInt`.

PR #289, opened 2022-06-02
StaticInt for n-first/last

Tokazama:


chriselrod:
I've approved, but in what way is this better than just dispatching to `first(x, Int(n))` and `last(x, Int(n))`?

Tokazama:
> I've approved, but in what way is this better than just dispatching to `first(x, Int(n))` and `last(x, Int(n))`?
> 
> I'd see the benefit if this preserved the static information, but it does not. Although it would let someone else define `getindex(::TheirVectorType, ::OptionallyStaticUnitRange)` overloads.

Do you mean that it only preserves static information as far as creating the index into `x` of `first(x, n)`? The reason I didn't just convert to `Int` was to support things things like `getindex(::TheirVectorType, ::OptionallyStaticUnitRange)`, but is there something better you had in mind?

chriselrod:
Yeah, this seems fine.

PR #300, opened 2022-06-06
Test Static 0.7

ChrisRackauckas:


Tokazama:
I dropped the `==` method for `Symbol` and `StaticSymbol` because it causes a bunch of invalidations. There's a [PR in base](https://github.com/JuliaLang/julia/pull/44678) to fix that but it hasn't had any movement lately. It would probably be easier for me to just change the tests here.

Tokazama:
I've fixed the dimension tests. There are still problems with `to_shape` but I should probably sit down and investigate potential invalidations before defining something there.

Tokazama:
Now that this is working we can get a good feel for the improvement on load time.

* master: 0.099550 seconds (127.35 k allocations: 7.828 MiB, 6.37% compilation time)

* PR: 0.050113 seconds (83.22 k allocations: 5.470 MiB, 10.80% compilation time)

chriselrod:
Can we merge this?

If we need additional changes, we can make another release, and the dependent packages can set that ArrayInterface version as their lower bound.

Tokazama:
I guess we need to rip off the band-aid eventually...

PR #301, opened 2022-06-06
fix non-reshapred reinterpretarray

chriselrod:
There was a regression introduced because `Base.ReinterpretArray`s aren't considered forwarding wrappers, which several methods were relying on.


Tokazama (approved):
Thanks sorry I missed this one

PR #302, opened 2022-06-06
OffsetArrays is forwarding wrapper

chriselrod:
Fixes some VectorizationBase failures

https://github.com/JuliaArrays/ArrayInterface.jl/runs/6760119793?check_suite_focus=true

Not sure how all these regressions got by us earlier?

### DIFF
--- a/lib/ArrayInterfaceOffsetArrays/Project.toml
+++ b/lib/ArrayInterfaceOffsetArrays/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterfaceOffsetArrays"
 uuid = "015c0d05-e682-4f19-8f0a-679ce4c54826"
-version = "0.1.2"
+version = "0.1.3"
 
 [deps]
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
--- a/lib/ArrayInterfaceOffsetArrays/src/ArrayInterfaceOffsetArrays.jl
+++ b/lib/ArrayInterfaceOffsetArrays/src/ArrayInterfaceOffsetArrays.jl
@@ -44,5 +44,6 @@ end
     d = ArrayInterface.to_dims(A, dim)
     OffsetArrays.IdOffsetRange(ArrayInterface.axes(parent(A), d), relative_offsets(A, d))
 end
+@inline ArrayInterface.is_forwarding_wrapper(::Type{<:OffsetArrays.OffsetArray}) = true
 
 end # module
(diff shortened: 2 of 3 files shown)


chriselrod:
Note there is still a VectorizationBase test failure brought on by recent ArrayInterface developments:
```julia
    [5] grouped_strided_pointer(A::Tuple{LinearAlgebra.Adjoint{Float64, Matrix{Float64}}, SizedWrapper{5, 6, Float64, LinearAlgebra.Adjoint{Float64, Matrix{Float64}}}, LinearAlgebra.Adjoint{Float64, Matrix{Float64}}}, #unused#::Val{(((1, 1), (3, 1)), ((1, 2), (2, 1)), ((2, 2), (3, 2)))})
      @ LayoutPointers ~/.julia/dev/LayoutPointers/src/grouped_strided_pointers.jl:62
```

Tokazama:
I'm not sure if we should consider OffsetArray a forwarding wrapper because it changes indices. Maybe the problem is with how `StrideIndex` is constructed

chriselrod:
Is CI using the actual PR when testing dependencies?
That would explain why VectorizationBase is still failing here, while it passed on the PR that actually broke it.

chriselrod:
> I'm not sure if we should consider OffsetArray a forwarding wrapper because it changes indices. Maybe the problem is with how `StrideIndex` is constructed

It would be good to have `default_to_forwarding_wrapper_except_when_we_explicitly_overload_otherwise`.
This is how the `parent_type` check was often used.
For many array types, most traits are forwarded, so it's nice to get those for free, while overloading only those that make that particular type special.

chriselrod:
Feel free to make a PR, otherwise I'll probably just upper bound ArrayInterface.

Tokazama:
We could just define this

```julia
ArrayInterface.stride_rank(T::Type{<:OffsetArray}) = stride_rank(parent_type(T))
ArrayInterface.contiguos_axis(T::Type{<:OffsetArray}) = contiguos_axis(parent_type(T))
```

There's probably a way to get `contiguous_axis` from `stride_rank` so that we don't have to worry about unwrapping anything.

chriselrod:
> We could just define this
> 
> ```julia
> ArrayInterface.stride_rank(T::Type{<:OffsetArray}) = stride_rank(parent_type(T))
> ArrayInterface.contiguos_axis(T::Type{<:OffsetArray}) = contiguos_axis(parent_type(T))
> ```
> 
> There's probably a way to get `contiguous_axis` from `stride_rank` so that we don't have to worry about unwrapping anything.

Still, something with that meaning would be useful, as otherwise people are going to have to implement the entire API.

chriselrod:
Closed in favor of #303

chriselrod:
Yeah, I think it makes sense to favor correctness, but I think it's likely the `is_forwarding_wrapper` is going to cause regressions in most packages that tried to implement the ArrayInterface.

Tokazama:
It's starting to look that way. We might need some sort of intermediate solution.

Issue #309, opened 2022-06-10
Wrong `first`  for `LazyAxis`


wangl-cc:
For a `x::LazyAxis{N}` whose first is not known, its `first` is defined as `offset(parent(x), N)`.

https://github.com/JuliaArrays/ArrayInterface.jl/blob/de94815f9bf24383d9121325d397614005686c94/src/axes.jl#L221-L227

However, the `parent(x)` is not its parent array, but `axes(x.parent, N)`

https://github.com/JuliaArrays/ArrayInterface.jl/blob/de94815f9bf24383d9121325d397614005686c94/src/axes.jl#L193

Thus if `N >= 2`, `first(x::LazyAxis{N}` will always `1`, because `ndims(axes(x.parent, N) = 1`.

This will break indexing for offset arrays
```julia
julia> using ArrayInterface, ArrayInterfaceOffsetArrays, OffsetArrays

julia> OA = OffsetArray(reshape(1:6, 2, 3), 3:4, 0:2)
2×3 OffsetArray(reshape(::UnitRange{Int64}, 2, 3), 3:4, 0:2) with eltype Int64 with indices 3:4×0:2:
 1  3  5
 2  4  6

julia> ArrayInterface.getindex(OA, :, :)
2×3 OffsetArray(::Matrix{Int64}, 3:4, 0:2) with eltype Int64 with indices 3:4×0:2:
 3  5  3539874564279314473
 4  6  4714751117537388841
```

Tokazama:
Fixed with #310

PR #310, opened 2022-06-10
Fix `to_index` for `CartesianIndices` and `first` for `LazyAxis`

### LINKED ISSUE #308: Incorrect  `getindex` for `CartesianIndices`
`getindex` for `CartesianIndices` whose indices are not start from `1` seems incorrect.

```julia
julia> A = reshape(1:6, 2, 3)
2×3 reshape(::UnitRange{Int64}, 2, 3) with eltype Int64:
 1  3  5
 2  4  6

julia> ArrayInterface.getindex(A, CartesianIndices((1:2, 2:3)))
2×2 Matrix{Int64}:
 1  3
 2  4

julia> ArrayInterface.getindex(A, 1:2, 2:3)
2×2 Matrix{Int64}:
 3  5
 4  6

julia> getindex(A, CartesianIndices((1:2, 2:3)))
2×2 Matrix{Int64}:
 3  5
 4  6
```
IMO, `ArrayInterface.getindex(A, CartesianIndices((1:2, 2:3)))` should be the same as `ArrayInterface.getindex(A, 1:2, 2:3)` which is the behavior of `Base.getindex`,
but it was treated as `getindex(A, 1:2, 1:2)`. 

It seems that `to_index` don't handle `CartesianIndices` properly:
https://github.com/JuliaArrays/ArrayInterface.jl/blob/e74a7aa619d67331d36476bcbff0bd29b0f31fa6/src/indexing.jl#L190

Returning `i.indices` should fix it:
```julia
julia> ArrayInterface.to_index(x, i::CartesianIndices) = i.indices

julia> ArrayInterface.getindex(A, CartesianIndices((1:2, 2:3)))
2×2 Matrix{Int64}:
 3  5
 4  6
```

wangl-cc:
Fix #308, #309

### DIFF
--- a/lib/ArrayInterfaceOffsetArrays/test/runtests.jl
+++ b/lib/ArrayInterfaceOffsetArrays/test/runtests.jl
@@ -11,6 +11,9 @@ Op = PermutedDimsArray(O,(3,1,2));
 @test @inferred(ArrayInterface.offsets(O)) === (4, 8, 11)
 @test @inferred(ArrayInterface.offsets(Op)) === (11, 4, 8)
 
+@test @inferred(ArrayInterface.to_indices(O, (:, :, :))) == (4:6, 8:11, 11:15)
+@test @inferred(ArrayInterface.to_indices(Op, (:, :, :))) == (11:15, 4:6, 8:11)
+
 @test @inferred(ArrayInterface.offsets((1,2,3))) === (StaticInt(1),)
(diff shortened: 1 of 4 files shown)


Tokazama (approved):
Thanks! I'll try to get some more exhaustive tests for `LazyAxis` in soon so we can ensure this sort of thing doesn't happen anymore.

PR #315, opened 2022-06-14
`IndicesInfo` for succinct gathering of info on indices.

Tokazama:
This tidies up the index traits into a single call, `IndicesInfo` and helps avoid directly indexing the `parameters` field within the generated method of `to_indices` (since access to that field directly is discouraged).

### DIFF
--- a/lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
+++ b/lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
@@ -576,9 +576,8 @@ IndicesInfo(@nospecialize x::Tuple) = IndicesInfo(typeof(x))
         push!(NS.args, :(ndims_shape($(T_i))))
         push!(IS.args, :(is_splat_index($(T_i))))
     end
-    :(IndicesInfo{$(NI),$(NS),$(IS)}())
+    Expr(:block, Expr(:meta, :inline), :(IndicesInfo{$(NI),$(NS),$(IS)}()))
 end
-
 
 """
     instances_do_not_alias(::Type{T}) -> Bool
--- a/lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
+++ b/lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
@@ -557,6 +557,13 @@ ndims_shape(@nospecialize T::Type{<:Number}) = 0
 ndims_shape(@nospecialize T::Type{<:AbstractArray}) = ndims(T)
 ndims_shape(x) = ndims_shape(typeof(x))
 
+"""
+    IndicesInfo(T::Type{<:Tuple}) -> IndicesInfo{NI,NS,IS}()
+
+Provides basic trait information for each index type in in the tuple `T`. `NI`, `NS`, and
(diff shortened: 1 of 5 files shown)


lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
```diff
@@ -562,6 +557,29 @@ ndims_shape(@nospecialize T::Type{<:Number}) = 0
...
+    NS = Expr(:tuple)
+    IS = Expr(:tuple)
+    for i in 1:fieldcount(T)
+        T_i = fieldtype(T, i)
+        push!(NI.args, :(ndims_index($(T_i))))
+        push!(NS.args, :(ndims_shape($(T_i))))
+        push!(IS.args, :(is_splat_index($(T_i))))
+    end
+    :(IndicesInfo{$(NI),$(NS),$(IS)}())
+end
```
chriselrod, line 580: Seems unfortunate that this sort of thing has to be a generated function.
If we have a lot of code that wants to do something like this, it may be worth defining a `typetail` to take a recursive approach.

src/indexing.jl
```diff
@@ -91,69 +81,97 @@ This implementation differs from that of `Base.to_indices` in the following ways
...
+            end
+        end
+
+        # 3. insert `to_index` calls
+        dim = 0
+        nndi = length(ndi)
+        for i in 1:nndi
+            if dim == N
+                # at this point each dimension has an index assigned to it and remaining indices
+                # are we set the ndims_index argument corresponding to it to zero so that it's
```
chriselrod, line 141: comments could use some rephrasing

chriselrod (approved):

lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
Tokazama: The generated function itself isn't costing very much in load time but if we end up needing to do the same thing a lot we could have a single generated `each_field(f, ::Type{<:Tuple})` to avoid defining too many of these. What we need is essentially `ntuple(i -> f(fieldtype(T, i)), Val{fieldcount(T)}())` but I've noticed constant propagation gives up pretty easily with `fieldtype`.

Either way, I should at least try being proactive about this before merging so I'll try a few things out before merging this.

chriselrod: While I do think latency is a major issue, I am more worried about const prop failures.
(And many latency problems do come from type instabilities.)

Tokazama: I tried using this code instead:
```julia
struct IndexInfo{NI,NS,IS} end
IndexInfo(x) = IndexInfo{ndims_index(x),ndims_shape(x),is_splat_index(x)}()

indices_info(@nospecialize x::Tuple) = indices_info(typeof(x))
function indices_info(::Type{T}) where {T<:Tuple}
    ntuple(i -> IndexInfo(fieldtype(T, i)), Val{fieldcount(T)}())
end
```

and inference was fine. I'm not completely confident it will hold up in practice because at the very least it will be used for `ArrayInterface._to_indices` which is already several steps away from the call site `ArrayInterface.getindex`.

If there's something else I'm missing I'm open to trying it out.

PR #316, opened 2022-06-17
Safe functional indexing

Tokazama:
This allows common comparison operators be parsed into indices (`<`, `<=`, `>`, `>=`). So instead of doing `x[3:end]` you can do `x[>=(3)]` (if `x` uses `ArrayInterface.getindex`. The main utility right now is that we can ensure that the indices are always inbounds using this syntax. I'm still working out the best way to communicate this info to `checkbounds` so that bounds checking is passed over (similar to how `:` becomes `Slice` which is known to be inbounds); but that will probably be more involved so I'm going to make that a separate effort.

### DIFF
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -36,6 +36,9 @@ import Compat
 
 n_of_x(::StaticInt{N}, x::X) where {N,X} = ntuple(Compat.Returns(x), Val{N}())
 
+_add1(@nospecialize x) = x + oneunit(x)
+_sub1(@nospecialize x) = x - oneunit(x)
+
 @generated function merge_tuple_type(::Type{X}, ::Type{Y}) where {X<:Tuple,Y<:Tuple}
     Tuple{X.parameters...,Y.parameters...}
 end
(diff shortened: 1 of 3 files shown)


Tokazama:
@wangl-cc, you're welcome to review this since it may be of interest to you.

wangl-cc:
If `checkbounds` is needed, I think `min` or `max` can be removed, like `x[begin:3]`

Besides, I think we can create some "functional range" like `getindex(x, 3:last)`, which is much similar to the syntax `x[3:end]`.

Tokazama:
The reason I used min and max was because those functions are returning all indices below or above the provided value. This is the same as if you filtered the axis with those functions. Maybe this needs to be clarified more in the docs.

We can't create a functional range like that because it would be type piracy.

wangl-cc:
> The reason I used min and max was because those functions are returning all indices below or above the provided value. This is the same as if you filtered the axis with those functions. Maybe this needs to be clarified more in the docs.

With this interpretation, I think keeping them always inbounds is OK.

> We can't create a functional range like that because it would be type piracy.

Maybe we can create new types as alternatives `begin` and `end` to avoid type piracy.

Tokazama:
There is a package for the last suggestion, so it's probably more intuitive to do something similar or provide support through that package. I'll remove the `firstindex` and `lastindex` ones for now.

PR #317, opened 2022-06-18
`SubCartesianIndices` to avoid allocating

wangl-cc:
This PR creates a new type `SubCartesianIndices` for `unsafe_get_collection(::CartesianIndices, ind)` to avoid allocating a new array. And a `SubCartesianIndices` can also be an index like `CartesianIndices`.

I also tried to create the `SubLinearIndices`. But `view(LA, ...)` may be the easiest way to avoid allocating, a `SubLinearIndices` may not be better. Besides, I have no idea about treat a `SubCartesianIndices` as an index. Any suggestion? @Tokazama

Besides, there are also some issue about `unsafe_get_collection(:: LinearIndices, ind)`
```julia
julia> A = reshape(1:12, (3, 4)); 

julia> LA = LinearIndices(A)
3×4 LinearIndices{2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}:
 1  4  7  10
 2  5  8  11
 3  6  9  12

julia> ArrayInterface.getindex(LA, static(1):2, static(1):2)
2×2 reshape(::LinearIndices{2, Tuple{ArrayInterface.OptionallyStaticUnitRange{StaticInt{1}, Int64}, ArrayInterface.OptionallyStaticUnitRange{StaticInt{1}, Int64}}}, 2, 2) with eltype Int64:
 1  3
 2  4

julia> LA[static(1):2, static(1):2]
2×2 Matrix{Int64}:
 1  4
 2  5
```
This means that we can only avoid allocating only if `size(LA, i) == length(inds[i])` for `i` in `1:(ndims(LA)-1)` currently.


Tokazama:
I wouldn't worry about `LinearIndices`. I'll have to take a closer look at the rest of the code later tonight or tomorrow.

wangl-cc:
I think there are some issue about diff coverage. Lines warned by codecov have been covered.

Tokazama:
It would be nice if we could return `NDIndex` so that a `StaticInt` index could be preserved in iteration since it will never change. It's harder to make that work though because each set of indices would need to construct a new element type.

wangl-cc:
@Tokazama Can we extend the `static_unsafe_getindex` as definition of `ArrayInterface.getindex` for ranges? It might be useful to preserve `StaticInt`.

Tokazama:
Maybe we shouldn't worry about forcing this to work with `CartesianIndices` and we just want our own type that functions as a more efficient replacement?

Tokazama:
> @Tokazama Can we extend the `static_unsafe_getindex` as definition of `ArrayInterface.getindex` for ranges? It might be useful to preserve `StaticInt`.

Could you give me an example of what you mean?

wangl-cc:
The `Base.getindex` will always returns a `Int`, but it will preserve it as a `StaticInt` when possible.
```julia
julia> (static(1):static(2))[static(1)]
1

julia> ArrayInterface.static_unsafe_getindex(static(1):static(2), static(1))
static(1)
```

wangl-cc:
> Maybe we shouldn't worry about forcing this to work with `CartesianIndices` and we just want our own type that functions as a more efficient replacement?

So we should't make it as return of `getindex` for `CartesianIndices`?

Tokazama:
> > Maybe we shouldn't worry about forcing this to work with `CartesianIndices` and we just want our own type that functions as a more efficient replacement?
> 
> So we should't make it as return of `getindex` for `CartesianIndices`?

Is the use case something like `view(A, i1, i2)[static(1), :]`?

Tokazama:
> > Maybe we shouldn't worry about forcing this to work with `CartesianIndices` and we just want our own type that functions as a more efficient replacement?
> 
> So we should't make it as return of `getindex` for `CartesianIndices`?

It seems like we might actually just want to replace `CartesianIndices` with this new type entirely so we don't have to hack our way around so many issues. If this is way more work than you want to commit to at this point that's totally fine. I just want to make sure that if we create a whole new array type it's not just another band-aid to a larger issue. Otherwise we're stuck maintaining it and developing a full solution here.

Tokazama:
> > It seems like we might actually just want to replace `CartesianIndices` with this new type entirely so we don't have to hack our way around so many issues. If this is way more work than you want to commit to at this point that's totally fine. I just want to make sure that if we create a whole new array type it's not just another band-aid to a larger issue. Otherwise we're stuck maintaining it _**and**_ developing a full solution here.
> 
> Got it, what we need might be not a `SubCartesianIndices` but a `NDIndices`.

Exactly. And we might need to move slowly on this or in small incremental PRs that clearly state this part is experimental and subject to change. It's one of those things that's hard to implement all at once or incrementally because it we have to consider a lot of things it will be used for.

Tokazama:
@wangl-cc , sorry this is moving slowly. I think everybody maintaining this package right now is trying to solve other more pressing issues right now, including myself. Hopefully we can come back to this in the near future.

Your contributions are always appreciated.

wangl-cc:
@Tokazama, thanks for your reply, we can continue this PR when you are free.

PR #319, opened 2022-06-22
Improve performance of recent function-index

Tokazama:
Previous versions were trying to use `firstindex` and `lastindex` to reference an axis but `first(axis)` and `last(axis)` is what we actually want and is faster.


chriselrod (approved):

Tokazama:
@ChrisRackauckas, I'm assuming those SciMLBase failures are unrelated?

ChrisRackauckas:
Passed in the second run. Thanks for letting me know, that was a goof on my end.

PR #325, opened 2022-06-24
Create ArrayInterfaceStaticArraysCore

ChrisRackauckas:
Now that there's StaticArraysCore, to have minimal dependencies I created ArrayInterfaceStaticArraysCore which only requires ArrayInterfaceCore and StaticArraysCore and implements all of the possible functions given those constraints (i.e. no static-ness), which covers all of the functions which are actually used by SciML downstream packages. Thus this will have a massive effect on downstream load times. As a result, a patch bump to ArrayInterfaceStaticArrays is done which just includes the core version so it's non-breaking.


ChrisRackauckas:
Heavily tested locally.

PR #327, opened 2022-06-28
New dimension mapping

Tokazama:
## Why do this and what are the objectives

The primary focus here is to replace the `to_parent_dims` and `from_parent_dims` methods which were only really being used for `SubArray`, `Adjoint`, `Transpose`, and `PermutedDimsArray`. The point of those methods was originally to avoid repeatedly doing verbose extraction of the permutations for those last three arrays and to consolidate the number of generated functions for mapping across the indices of `SubArray`. However, there are some problems with `SubArray` that require more than simple dimension to dimension mapping.

The main goal here is to have something that actually produces an accurate dimension mapping for `SubArray`. A sort of secondary goal is to reduce many of the generated functions we've needed to support `SubArray`. Inference is working on local tests (Julia v1.6, v1.7, v1.8,) but seems to be failing CI on v1.6 for some reason. I guess I'll have to do some trial and error there through multiple PRs.

This is _not_ a proposal for a formal interface at this time.
In fact, I'm just removing the formal dimension mapping that `to_parent_dims` and `from_parent_dims` composed, because it just wasn't working.
I've tried to include a lengthy explanation of how the internals work in the code comments so everyone knows what's going on as we move forward.

## Points of Review

@chriselrod, it appears this is still safe for the LV stuff (which was my main concern moving forward with a release). I would think it's possible to replace the generated function for `dense_dims` that `SubArray` requires but I haven't had time to figure out what's going on with that code there yet. I welcome any ideas you have to simplify that but I probably won't have time to figure it out and replace it with something that isn't generated _and_ is performant.
Other than that, I've checked that the methods I'm removing aren't used in any currently used version of `ArrayInterface` in downstream packaging.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -74,6 +74,7 @@ ArrayInterface.axes_types
 ArrayInterface.broadcast_axis
 ArrayInterface.deleteat
 ArrayInterface.dense_dims
+ArrayInterface.from_parent_dims
 ArrayInterface.getindex
 ArrayInterface.indices
 ArrayInterface.insert
@@ -88,6 +89,7 @@ ArrayInterface.to_axis
 ArrayInterface.to_dims
 ArrayInterface.to_index
 ArrayInterface.to_indices
+ArrayInterface.to_parent_dims
 ArrayInterface.unsafe_reconstruct
 ```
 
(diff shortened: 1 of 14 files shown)


chriselrod (approved):
I don't think removing `@generated` is a high priority/it can be done in a later PR.

PR #328, opened 2022-06-30
`index_labels` method and `IndexLabel` type indexing

Tokazama:
Successor to #260. Implements basic tools for accessing multidimensional keys and fancy indexing.

The goal is to have a common backend for some of the clearly overlapping traits/method that go into `AxisArray`, `DimensionalData`, and `AxisKeys`. This should also help us get closer to helping other packages provide support for multiple fancy array types without explicitly depending on them.

@johnnychen94 , @rafaqz , @mcabbott may be interested.

### DIFF
--- a/lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
+++ b/lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
@@ -652,7 +652,7 @@ is returned.
 ndims_index(::Type{<:Base.AbstractCartesianIndex{N}}) where {N} = N
 # preserve CartesianIndices{0} as they consume a dimension.
 ndims_index(::Type{CartesianIndices{0,Tuple{}}}) = 1
-ndims_index(@nospecialize T::Type{<:Union{Number,Key,Symbol,AbstractString}}) = 1
+ndims_index(@nospecialize T::Type{<:Union{Number,Key,Symbol,AbstractString,AbstractChar}}) = 1
 ndims_index(@nospecialize T::Type{<:AbstractArray{Bool}}) = ndims(T)
 ndims_index(@nospecialize T::Type{<:AbstractArray}) = ndims_index(eltype(T))
 ndims_index(@nospecialize T::Type{<:Base.LogicalIndex}) = ndims(fieldtype(T, :mask))
@@ -668,7 +668,7 @@ indexing with an instance of `I`.
 ndims_shape(T::DataType) = ndims_index(T)
 ndims_shape(::Type{Colon}) = 1
 ndims_shape(@nospecialize T::Type{<:CartesianIndices}) = ndims(T)
-ndims_shape(@nospecialize T::Type{<:Union{Number,Base.AbstractCartesianIndex,Key,Symbol,AbstractString}}) = 0
+ndims_shape(@nospecialize T::Type{<:Union{Number,Base.AbstractCartesianIndex,Key,Symbol,AbstractString,AbstractChar}}) = 0
 ndims_shape(@nospecialize T::Type{<:AbstractArray{Bool}}) = 1
 ndims_shape(@nospecialize T::Type{<:AbstractArray}) = ndims(T)
 ndims_shape(x) = ndims_shape(typeof(x))
--- a/lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
+++ b/lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
@@ -561,12 +561,12 @@ Base.@propagate_inbounds function Base.getindex(ind::TridiagonalIndex, i::Int)
 end
 
 """
-    Key(key)
+    Label(label)
 
-A type that clearly communicates that `key` refers to a key-index mapping.
+A type that clearly communicates that `label` refers to a key-index mapping.
 """
-struct Key{K} <: ArrayIndex{0}
-    key::K
+struct Label{L} <: ArrayIndex{0}
+    label::L
 end
 
 _cartesian_index(i::Tuple{Vararg{Int}}) = CartesianIndex(i)
@@ -652,7 +652,7 @@ is returned.
 ndims_index(::Type{<:Base.AbstractCartesianIndex{N}}) where {N} = N
 # preserve CartesianIndices{0} as they consume a dimension.
 ndims_index(::Type{CartesianIndices{0,Tuple{}}}) = 1
-ndims_index(@nospecialize T::Type{<:Union{Number,Key,Symbol,AbstractString,AbstractChar}}) = 1
+ndims_index(@nospecialize T::Type{<:Union{Number,Label,Symbol,AbstractString,AbstractChar}}) = 1
 ndims_index(@nospecialize T::Type{<:AbstractArray{Bool}}) = ndims(T)
 ndims_index(@nospecialize T::Type{<:AbstractArray}) = ndims_index(eltype(T))
 ndims_index(@nospecialize T::Type{<:Base.LogicalIndex}) = ndims(fieldtype(T, :mask))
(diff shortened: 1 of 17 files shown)


Tokazama:
After a month of dedicated [bike-shedding on discourse](https://discourse.julialang.org/t/commit-to-a-common-syntax-for-accessing-additional-index-mapping-information-on-axes/84988), I think this is ready for review.

I'd like to have relevant package maintainers review the name of the function @rafaqz , @mcabbott, and someone from AxisArrays (maybe @johnnychen94). I'd also like an okay on the syntax from @ChrisRackauckas because we're using the term "label" here and I want to ensure that this isn't problematic for LabelledArrays.jl.

ChrisRackauckas:
I don't care about the name

lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
```diff
@@ -616,6 +616,15 @@ Base.@propagate_inbounds function Base.getindex(ind::TridiagonalIndex, i::Int)
     end
 end
 
+"""
+    Label(label)
+
+A type that clearly communicates that `label` refers to a key-index mapping.
```
nalimilan, line 622: This sounds like a very generic name, especially to clearly communicates that `label` refers to a key-index mapping". ;-) Maybe `IndexLabel`?

src/axes.jl
```diff
@@ -248,3 +247,64 @@ lazy_axes(x::AbstractRange, ::StaticInt{1}) = Base.axes1(x)
 lazy_axes(x, ::Colon) = LazyAxis{:}(x)
 lazy_axes(x, ::StaticInt{dim}) where {dim} = ndims(x) < dim ? SOneTo{1}() : LazyAxis{dim}(x)
 @inline lazy_axes(x, dims::Tuple) = map(Base.Fix1(lazy_axes, x), dims)
+
+"""
+    index_labels(x)
+    index_labels(x, dim)
+
+Returns a tuple of labels assigned to each axis or a collection of labels corresponding to
+axis `dim` of `x`. Default is to simply return `map(keys, axes(x))`.
```
nalimilan, line 256: Is it really a good idea to have the default return `keys`? I'd say better return `nothing` so that the caller knows that no labels are set (it's easy to call `keys` if needed). Otherwise there's no way to know whether labels are present or not.

src/axes.jl
```diff
@@ -248,3 +247,64 @@ lazy_axes(x::AbstractRange, ::StaticInt{1}) = Base.axes1(x)
...
+"""
+index_labels(x, dim)  = index_labels(x, to_dims(x, dim))
```
nalimilan, line 258: Shouldn't the function live in ArrayInterfaceCore so that existing "named array" packages can overload it? BTW, it would be good to ping their authors to ensure they would all be OK with the API, otherwise it won't make a lot of sense.

lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
Tokazama: agreed

src/axes.jl
Tokazama: Existing packages can overload it from `ArrayInterface`. If you're referring to supporting named dimensions like NamedDims.jl, then they define `ArrayInterface.dimnames` and `to_dims` maps to the appropriate dimension so they don't have to overload every method with a dim argument.

src/axes.jl
Tokazama: I should probably have changed that back to just the axis after all the discussion around naming this. My thinking was that it would be odd if we returned something that didn't "label" the indexes at all because we are assuming that is what's returned.

Returning `nothing` would require that every method that acts on index labels would require and extra step to check if it can actually look up values in it. Not an insurmountable issue, but it would certainly be tedious.

Throwing an error when this isn't defined also isn't an option because things like a vector transpose will inevitably have a dimension without labels.

src/axes.jl
nalimilan: ArrayInterface is relatively heavy, which is why ArrayInterfaceCore was created IIUC. I guess it's up to package authors to say whether a dependency on ArrayInterface is acceptable for them or not.

src/axes.jl
nalimilan: If we don't return `nothing` when no labels are set, then we should add a separate function called e.g. `has_index_labels`, a bit like we did in the Metadata PR with `hasmetadata` and `metadata`.

src/axes.jl
Tokazama: That was more of an issue before we went through all the trouble to fix invalidations due to `StaticInt` earlier this year. That doesn't mean we can't improve the situation. We are actively trying to move matured functionality into base where appropriate (see #340). I regularly review the code here in an effort to eliminate problematic code that still exists (e.g., redundancies, generated functions, etc,). For example, once we know how this PR is going to look I can finally finish an effort to consolidate a lot of "indexing.jl" by overloading base methods instead of reimplementing a lot of what's in base already.

src/axes.jl
Tokazama: That's definitely doable and is probably a good method to have irregardless of what the default we choose.

I'm not against returning something that clearly shows no labels are defined. Perhaps I'm being too strict with the rule that the return has the same indices as their corresponding axis, even for default values. I assume people will at least expect to be able to do things like `findfirst(==(label), index_labels(A, dim))`, in which case we would probably still want a more informative error.

rafaqz: I would also prefer `nothing`

Tokazama: I think the issues I mentioned can be avoided with documentation and tooling if we decide to return nothing. But AxisKeys.jl uses a similar approach to what's implemented here so it would probably be good if @mcabbott let us know if we can make this work for AxisKeys

src/axes.jl
Tokazama: Also, it's a bit of a chicken and egg issue. We use `StaticSymbol` for dimension names known at compile time so that we can use them as a point of reference in an inferrible way. It's pretty difficult to do this only relying on constant propagation (demonstrated that with static sizes here https://github.com/JuliaLang/julia/pull/44538#issuecomment-1229205583).

If someone has a reliable solution I'm open to it. I've been trying to actively address this and related issues for years

src/axes.jl
rafaqz: The issue for DimensionalData.jl is to be able to not propagate the index labels when there are none, such as in`cat`, where joining multiple `Base.OneTo` ranges from `keys` is really not what you want to do.

In DD I use an explicit `NoLookup` object to solve this - it works correctly in both `findfirst` and `cat`. The problem for me will be interop with other types, where there needs to be a way of knowing if there are real index labels associated with an axis, not just `keys`.

Probably a `has_index_labels` trait would work for this too, maybe it is better than returning `nothing`.

Tokazama: Good point. A default without a clearly know type may cause more problems than it solves

Tokazama: > In DD I use an explicit NoLookup object to solve this - it works correctly in both findfirst and cat. The problem for me will be interop with other array types, where there needs to be a way of knowing if there are real index labels associated with an axis, not just keys.

What about returning an `UndefLabels`?

src/axes.jl
Tokazama: @oxinabox is it still a concern depending on ArrayInterface at this point?

oxinabox: Now that ArrayInterface doesn't like Requires.jl a billion packages, I am much more comfortable depending upon it for NamedDims.jl

nalimilan: My suggestion was just to put empty function definitions in ArrayInterfaceCore (like we do with DataAPI and StatsAPI). That doesn't prevent keeping fallback method definitions in ArrayInterface as this PR does. But packages that don't want to use fallback definitions are still able to overload the functions by depending only on ArrayInterfaceCore.

Tokazama: We've been trying to avoid situations where the behavior of a method is different if ArrayInterface is loaded vs just ArrayInterfaceCore.

src/axes.jl
Tokazama: > Is it really worth defining a special type just for that? What would be the advantage over Nothing?

I'm just trying to figure out some generic representation of labels not being defined that can pass between methods. If we do `vcat(a, b)` but one doesn't have index labels, then we can't do `vcat(index_labels(a, 1), index_labels(b, 1))`. Even if we define something that doesn't error in this situation (e.g., `vcat_index_labels(::AbstractVector, ::Nothing)`), how does this produce a new vector of index labels for the newly generated array? I figured `UndefLabel` could represent every index without a label.

@rafaqz how are you handling this situation?

Tokazama: @rafaqz, it looks like `NoLookup` is basically the same thing as what I'm describing. It wraps the non-labeled indices.

Tokazama: Also, do we want to have `has_index_labels(T::Type)` or `has_index_labels(x) = any(are_labels, index_labels)`?

Tokazama:
@mcabbott, would you mind providing some input if you have time?

src/axes.jl
rafaqz: Sorry I missed this earlier. If your `UndefLabels` is essentially my `NoLookup`, that should be fine.

ChrisRackauckas:

Issue #332, opened 2022-07-14
Storage Order


Wimmerer:
This may already be contained in ArrayInterface somewhere, but I haven't yet found it. It would be great if we could have a trait that indicates the storage order of an array here. Something like `ColMajor`, `RowMajor`, and a fallback `RuntimeOrder` and/or `UnknownOrder`.

Tokazama:
We have `stride_rank` and `is_column_major`

PR #334, opened 2022-07-24
Make `stride_rank` public and document

Tokazama:
We have the tooling to in place to solve #332. I think JuliaSIMD uses `stride_rank` too. So I figured it would make sense to just make this a public function and document it. @chriselrod, I'm pretty sure you were the original author on this so if there's something else to add to the documentation that would be welcomed.


ChrisRackauckas:

PR #342, opened 2022-09-02
Cleanup indices info + better docs

Tokazama:


Tokazama:
@ChrisRackauckas , are these errors expected right now https://github.com/JuliaArrays/ArrayInterface.jl/runs/8166027000?check_suite_focus=true

ChrisRackauckas:
Yes, sorry about that. That failure is unrelated and should disappear later today.

PR #343, opened 2022-09-02
`@inline` StrideIndex

chriselrod:
@ffreyer you can try this

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterface"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "6.0.22"
+version = "6.0.23"
 
 [deps]
 ArrayInterfaceCore = "30b0a656-2188-435a-8636-2ec0e6a096e2"
--- a/lib/ArrayInterfaceCore/Project.toml
+++ b/lib/ArrayInterfaceCore/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterfaceCore"
 uuid = "30b0a656-2188-435a-8636-2ec0e6a096e2"
-version = "0.1.17"
+version = "0.1.18"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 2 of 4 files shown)


ffreyer:
Doesn't seem to help
```julia
julia> @benchmark reflectorApply!($x, $τ, $y)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  46.028 μs … 210.703 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     49.761 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   52.597 μs ±  11.745 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▇▇███▇▅▃▁    ▁▂▃▃▃▃▂          ▂▂▁                            ▃
  ███████████▆▅███████▇▅▆▆▅▄█▇█████▇▅▅▅▅▅▅▆▇▇▇█▆▆▆▃▄▃▁▁▃▄▆▃▆▆▅ █
  46 μs         Histogram: log(frequency) by time       105 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.

(@v1.7) pkg> st ArrayInterface
      Status `~/.julia/environments/v1.7/Project.toml`
  [4fba245c] ArrayInterface v6.0.22 `https://github.com/JuliaArrays/ArrayInterface.jl.git#inlinestrideindex`
```

Tokazama (approved):

Tokazama:
We might need a more aggressive fix for for previous versions
```julia
function map_tuple_type(f::F, ::Type{T}) where {F,T<:Tuple}
    if @generated
        t = Expr(:tuple)
        for i in 1:fieldcount(T)
            push!(t.args, :(f($(fieldtype(T, i)))))
        end
        Expr(:block, Expr(:meta, :inline), t)
    else
        Tuple(f(fieldtype(T, i)) for i in 1:fieldcount(T))
    end
end
```

chriselrod:
> We might need a more aggressive fix for for previous versions

You're welcome to take over this PR, or we can merge it and you follow up with another?

I was waiting on integration tests to merge, but this PR seems like it should be harmless w/ respect to semantics.

Tokazama:
> > We might need a more aggressive fix for for previous versions
> 
> You're welcome to take over this PR, or we can merge it and you follow up with another?
> 
> I was waiting on integration tests to merge, but this PR seems like it should be harmless w/ respect to semantics.

Sounds good. I put in the old generated methods. I'll wait for some tests to pass before merging but your probably right that we don't need to wait for all of the integration tests.

PR #345, opened 2022-09-03
Overload `Base.to_index` (instead of using our own)

Tokazama:
Initial effort to get rid of our own internal definitions of `ArrayInterface.to_index` using new `AxisIndex` type for overloading with any array type (e.g., `Base.to_index(::Any, ::AxisIndex)`. `AxisIndex` contains all the information to retrieve the appropriate axes for a give index passed through `to_indices`. This also lets us dispatch on `Base.to_indices(A, ::Tuple{Vararg{<:ArrayIndex}})` now, so we are pretty close to eliminating the need for any unique methods on our part here. This implementation also doesn't cause any additional invalidations.

Initial benchmarks seemed to be good but I'm trying to balance excess method specialization with performance, which is likely to be highly dependent on Julia versions.

### DIFF
--- a/src/ArrayInterface.jl
+++ b/src/ArrayInterface.jl
@@ -29,7 +29,7 @@ using IfElse
 
 using Base.Cartesian
 using Base: @propagate_inbounds, tail, OneTo, LogicalIndex, Slice, ReinterpretArray,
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:

PR #346, opened 2022-09-08
 Fixing lu_instance doesnt work on julia-1.8 

TheZombie1999:
With the new release of Julia 1.8 the public interface of CUDA.CUSOLVER changed.
This requires that the implementation of the lu_instance function has to change.


ChrisRackauckas (approved):

ChrisRackauckas:
I added a quick test. Thanks!

TheZombie1999 (commented):
That test doesn't really work. Because in Julia 1.8 lu_instance doesn't return a CuQR struct any more.

TheZombie1999:
I created a new pull request. The return type is a QR struct. from linear algebra. 

My test simply looks wether the function throws an error or not.

PR #352, opened 2022-10-01
Rework `dimnames`

Tokazama:
We've had support for `dimnames` for a [long time](https://github.com/JuliaArrays/ArrayInterface.jl/releases/tag/v2.14.0) but no one has fully adopted it. I thought people would be on board with it once we improved load times (getting rid of `Requires`, fixing a bunch of invalidations). However, there's still resistance to using things like `StaticSymbol` and depending on `ArrayInterface` (instead of `ArrayInterfaceCore`). I figured I could start making PRs that change other people's packages or accept that after nearly two years of having this feature, people just don't like the way it's implemented.

This implementation is as barebones as I could get it, removing `known_dimnames` and the ability to access `dimnames` nested within other arrays.

@oxinabox and @rafaqz, let me know what you think.


rafaqz:
Don't take any lack of implementation as any kind of judgement, it's just lack of time.

I haven't adopted much in ArrayInterface essentially because I'm doing a PhD and maintain too many other packages. Work on DimensionalData.jl is mostly bugfixes, Rasters.jl needs, and direct feature requests. 

I was also worried about the compile time for a while, but good to see that is getting sorted.

Tokazama:
> Don't take any lack of implementation as any kind of judgement, it's just lack of time.
> 
> I haven't adopted much in ArrayInterface essentially because I'm doing a PhD and maintain too many other packages. Work on DimensionalData.jl is mostly bugfixes, Rasters.jl needs, and direct feature requests.
> 
> I was also worried about the compile time for a while, but good to see that is getting sorted.

A [quick search](https://juliahub.com/ui/Search?q=has_dimnames&type=code of package code) shows that literally no package has adopted this. I understand that people are busy working on other things, but at this point I think it's more likely that there are some flaws in the current implementation hindering adoption.

rafaqz:
I'm not sure. I would guess more that ArrayInterface is a large ambitious project and most of us do not really understand the capabilities yet.

Tokazama:
> There may also be some chicken/egg problems where no one implements something just because no one is implementing it, and it only makes sense to have at the collective level.

I could see that. One benefit to the lack of adoption is that we can review this and make changes without breaking any code.

Tokazama:
I think I might have found a good solution here with the latest implementation of `Names`. This may solve several issues with the original implementation of `dimnames`.

The original implementation used `StaticSymbol` in order to overcome issues with constant propagation and have a common type for passing around statically known dimension names. On top of that we wanted it to be something that end users didn't _have_ to know about (similar to how most people don't need to know about `Val`). I also thought adoption would be easier if I provided default propagation and inference of dimension names between nested layers of arrays. Here's what I think ended up happening:

* People really like the pattern of `f(x) = wrap(dimnames(x), f(parent(x))`, instead of trusting that dimension names won't disappear unexpectedly. Best case scenario, we had a bunch of extra code here that simply wasn't doing anything but increasing load times. Worst case scenario, users saw this internal implementation and decided it was too complex to figure out.
* `StaticSymbol` ended up not being functionally equivalent to `Symbol`, forcing people to learn how to manage some awkward situations. This is mostly because we can't even define `==(::StaticSymbol, ::Symbol)` without an enormous amount of invalidations. Simple things like `dimnames(x)[dim] == :x` ended up needing to be rewritten as `dynamic(dimnames(x)[dim]) == :x` to ensure it worked properly. This wasn't how things were originally designed, but it's where we ended up.
* The issue of a common storage type wasn't fully solved because `StaticSymbol` doesn't represent a collection of unique names (except `:_`), like we all originally wanted. Which leaves us wanting something similar but not the same as a `NamedTuple`.

I haven't implemented docs or tests for `Names` yet, but it functions a lot like if you were interacting with the keys of `NamedTuple` but it permitted multiple `:_` values. Finding symbols is known at compile time, without using any generated functions (basically copying patterns from "Base/namedtuple.jl"). With a little extra work this could probably be functionally equivalent to `NamedTuple` (using something like `index_labels` on a tuple based vector).

ChrisRackauckas:

PR #353, opened 2022-10-10
Add `undefmatrix` (similar to `zeromatrix`)

oscardssmith:
This is potentially much faster since it doesn't have to initialize values. For constructing a 1000x1000 `Matrix{Float64}` I benchmark this to be 10x faster which is very important because in cases like an autoswitching solver, you often want to preallocate a jacobian matrix, but might never use it if the equation isn't stiff.

### DIFF
--- a/lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
+++ b/lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
@@ -513,6 +513,27 @@ function zeromatrix(u::Array{T}) where {T}
 end
 
 """
+    undefmatrix(u::AbstractVector)
+
+Creates the matrix version of `u` with possibly undefined values. Note that this is unique because
+`similar(u,length(u),length(u))` returns a mutable type, so it is not type-matching,
+while `fill(zero(eltype(u)),length(u),length(u))` doesn't match the array type,
+i.e., you'll get a CPU array from a GPU array. The generic fallback is
+`u .* u'`, which works on a surprising number of types, but can be broken
+with weird (recursive) broadcast overloads. For higher-order tensors, this
+returns the matrix linear operator type which acts on the `vec` of the array.
+"""
+function undefmatrix(u)
+    x = safevec(u)
+    x .* x'
+end
+
+# Reduces compile time burdens
+function undefematrix(u::Array{T}) where {T}
+    out = Matrix{T}(undef, length(u), length(u))
+end
+                                                                                                                                    
+"""
     restructure(x,y)
 
 Restructures the object `y` into a shape of `x`, keeping its values intact. For
--- a/lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
+++ b/lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
@@ -524,13 +524,7 @@ with weird (recursive) broadcast overloads. For higher-order tensors, this
 returns the matrix linear operator type which acts on the `vec` of the array.
 """
 function undefmatrix(u)
-    x = safevec(u)
-    x .* x'
-end
-
-# Reduces compile time burdens
-function undefematrix(u::Array{T}) where {T}
-    out = Matrix{T}(undef, length(u), length(u))
+    similar(u, length(u), length(u))
 end
                                                                                                                                     
 """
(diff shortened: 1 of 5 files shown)


lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
```diff
@@ -512,6 +512,27 @@ function zeromatrix(u::Array{T}) where {T}
...
+"""
+function undefmatrix(u)
+    x = safevec(u)
+    x .* x'
+end
+
+# Reduces compile time burdens
+function undefematrix(u::Array{T}) where {T}
+    out = Matrix{T}(undef, length(u), length(u))
+end
```
chriselrod, line 534:
```suggestion
function undefmatrix(u)
    similar(u, length(u), length(u))
end
```

oscardssmith:
Should we merge this (so i can make the PR that adds this to ODE solvers)?

ChrisRackauckas:
This doesn't get the type correct for static arrays, and it doesn't have tests.

oscardssmith:
What should I be testing here? Array, CuArray, SArray?

oscardssmith:
Ah the first version worked with static array and sparse (and the performance difference is only there for large values anyway).

oscardssmith:
I've added tests for base types. The StaticArrays issue is arguably their own fault. They haven't defined a form of `similar` which it really feels like is their job.

chriselrod:
> I've added tests for base types. The StaticArrays issue is arguably their own fault. They haven't defined a form of `similar` which it really feels like is their job.

The problem is probably the call to `ArrayInterface.length`, which either is overloaded to return `StaticInt`s, or returns regular `Int`s.
If it is the latter, they're regular `Int`s and thus it'd risk type instabilities to have `similar(::Type{<:StaticArray}, ::Int, ::Int)` return a `StaticArray`.
If the former, yeah, it's probably `StaticArray`'s or `ArrayInterfaceStaticArray`'s job to define this overload.

ChrisRackauckas:
`similar` is defined in Base to return a mutable array.

```julia
help?> similar
search: similar

  similar(array, [element_type=eltype(array)], [dims=size(array)])

  Create an uninitialized mutable array with the given element type and size, based upon the given source array. The       
  second and third arguments are both optional, defaulting to the given array's eltype and size. The dimensions may be
  specified either as a single tuple argument or as a series of integer arguments.

  Custom AbstractArray subtypes may choose which specific array type is best-suited to return for the given element type   
  and dimensionality. If they do not specialize this method, the default is an Array{element_type}(undef, dims...).        

  For example, similar(1:10, 1, 4) returns an uninitialized Array{Int,2} since ranges are neither mutable nor support 2    
  dimensions:
```

chriselrod:
By `StaticArray` I meant the abstract type for which `MArray <: StaticArray`, not the immutable `SArray`.
Creating an undef `SArray` is useless.

Creating an `MArray` would be type unstable given `Int` args, while `StaticInt` would require a package to implement it.
It'd be type piracy if `ArrayInterfaceStaticArrays` does it, but that's probably okay.

ChrisRackauckas:
Add it to the docs

ChrisRackauckas:
Another test failure.

lib/ArrayInterfaceCore/test/runtests.jl
```diff
@@ -1,5 +1,5 @@
 using ArrayInterfaceCore
-using ArrayInterfaceCore: zeromatrix
+using ArrayInterfaceCore: zeromatrix, zeromatrix
```
ChrisRackauckas, line 2:
```suggestion
using ArrayInterfaceCore: zeromatrix, undefmatrix
```

oscardssmith: oops

ChrisRackauckas:
Bro.

oscardssmith:
fair. This time I've actually guaranteed that tests pass on my computer.

ChrisRackauckas (approved):

lib/ArrayInterfaceStaticArrays/src/ArrayInterfaceStaticArrays.jl
```diff
@@ -9,6 +9,14 @@ import ArrayInterfaceStaticArraysCore
...
 const CanonicalInt = Union{Int,StaticInt}
 
+function ArrayInterface.undefmatrix(::MArray{S, T, N, L}) where {S, T, N, L}
+    return MMatrix{L, L, T, L*L}(undef)
+end
+# SArray doesn't have an undef constructor and is going to be small enough that this is fine.
+function ArrayInterface.undefmatrix(s::SArray)
+    v = vec(s)
+    return v.*v'
+end
```
ChrisRackauckas, line 19: These could probably be staticarrayscore?

PR #354, opened 2022-10-11
Trait methods for `allunique` and `issorted`.

Tokazama:
In some cases, `allunique(::T)` and `issorted(::T)` will return `true` for all instances of `T`. In such cases `T` usually is overloaded for these and returns `true` with no operation, but in other cases this has to be checked at run time. This makes `issorted` and `allunique` poor methods for programmatic choices in code. For example, if we want to find a value we may end up iterating over the entire collection just to see if we can use `searchsortedfirst`.

This introduces trait method counterparts to `allunique` and `issorted` that only return `true` when this can be guaranteed from the type. ~~These are currently called `known_allunique` and `known_issorted` but I'm open to whatever syntax people prefer.~~

Changed the names to `is_set_like` and `ensures_sorted`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterface"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "6.0.23"
+version = "6.0.24"
 
 [deps]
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
Update to the latest version?

Tokazama:
Superseded by #386

PR #357, opened 2022-10-13
move `undefmatrix` overload to `ArrayInterfaceStaticarraysCore`

oscardssmith:
As requested by @ChrisRackauckas.


oscardssmith:
The downstream CI errros seem to be very unrelated (it appears someone changed the error messages without updating tests).

PR #362, opened 2022-10-30
Implement `all_assigned`

Tokazama:
`all_assigned(x)` performs `isassigned(x, i)` at each index for `x`. This implementation is very simple but may not be immediately obvious because it requires iterating over the indices _not_ the values.

### DIFF
--- a/lib/ArrayInterfaceCore/Project.toml
+++ b/lib/ArrayInterfaceCore/Project.toml
@@ -1,6 +1,6 @@
 name = "ArrayInterfaceCore"
 uuid = "30b0a656-2188-435a-8636-2ec0e6a096e2"
-version = "0.1.24"
+version = "0.1.25"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
+++ b/lib/ArrayInterfaceCore/src/ArrayInterfaceCore.jl
@@ -318,6 +318,18 @@ function Base.setindex(x::AbstractMatrix, v, i::Int, j::Int)
 end
 
 """
+    all_assigned(x) -> Bool
+
+Return `true` if `isassigned` is `true` at all indices of `x`.
+"""
+function all_assigned(x)
+    for i in eachindex(x)
+        @inbounds(isassigned(x, i)) || return false
+    end
+    return true
+end
+
+"""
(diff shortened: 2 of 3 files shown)


ChrisRackauckas:
What needs to extend this?

Tokazama:
My current use case is supporting the `:nothrow` effect.
Bounds checking isn't sufficient for this because an unassigned index will still throw an `UndefRefError`.

```julia
unsafe_getindex1(x, i::Int) = @inbounds(x[i])
Base.@assume_effects :nothrow unsafe_getindex2(x, i::Int) = @inbounds(x[i])
```

In case like `Vector` it makes sense that the compiler would struggle to figure this out
```julia
julia> Core.Compiler.infer_effects(unsafe_getindex1, Tuple{Vector{Int},Int})
(?c,+e,?n,+t)

julia> Core.Compiler.infer_effects(unsafe_getindex2, Tuple{Vector{Int},Int})
(?c,+e,+n,+t)
```

But even cases where we can't have an undefined error and in bounds should be sufficient, the compiler still doesn't know there won't be an error.
```julia
julia> Core.Compiler.infer_effects(unsafe_getindex1, Tuple{Tuple{Int},Int})
(?c,+e,?n,+t)

julia> Core.Compiler.infer_effects(unsafe_getindex2, Tuple{Tuple{Int},Int})
(?c,+e,+n,+t)
```

So any array whose inbound elements are guaranteed to be assigned could benefit from extending this.

Tokazama:
The previous example was just demonstrating that even if we know something is inbounds the compiler often can't ensure an error isn't going to be thrown. Assuring that all indices are defined allows us to make some assumptions that can speed things up sometimes

```julia
function find_first1(v, x)
    @assert ArrayInterfaceCore.all_assigned(x)
    i = unsafe_find_first(v, x)
    i === 0 ? nothing : i
end
Base.@assume_effects :terminates_locally :nothrow function unsafe_find_first(v, x)
    for i in eachindex(x)
        isequal(@inbounds(x[i]), v) && return i
    end
    return 0
end
function find_first2(v, x)
    for i in eachindex(x)
        isequal(@inbounds(x[i]), v) && return i
    end
    return 0
end
```

```julia
julia> let x = rand(100);
       @btime find_first1($(x[end]), $x)
       @btime find_first2($(x[end]), $x)
       end
  121.440 ns (0 allocations: 0 bytes)
  155.948 ns (0 allocations: 0 bytes)
```

ChrisRackauckas:
Update to the latest version?

PR #364, opened 2022-11-05
Add missing `parent_type` definitions

Tokazama:
This just adds `Hermitian` with the `Symmetric`'s `parent_type` definition.


Tokazama:
Test failures are from #305 since this is just adding something that should've been done already.

PR #366, opened 2022-11-07
Remove range types

Tokazama:
This was type piracy after the move of `StaticInt` to Static.jl.
Pending https://github.com/SciML/Static.jl/pull/88, this PR should solve that issue without any breaking changes for packages depending on ArrayInterface.jl.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,7 @@ Static = "aedffcd0-7271-4cad-89d0-dc628f76c6d3"
 ArrayInterfaceCore = "0.1.3"
 Compat = "3, 4"
 IfElse = "0.1"
-Static = "0.7"
+Static = "0.8"
 julia = "1.6"
 
 [extras]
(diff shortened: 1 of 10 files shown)


Tokazama:
@chriselrod , I think the only issue you might run into with this one is that `_pick_range` is gone, so I put a deprecation in for that so it would be mildly annoying instead of breaking.

chriselrod:
> @chriselrod , I think the only issue you might run into with this one is that `_pick_range` is gone, so I put a deprecation in for that so it would be mildly annoying instead of breaking.

StrideArrays.jl uses it, but none of the other repos that I saw.
It also depends on Static.jl, so the same PR bumping that can remove the use of `ArrayInterface._pick_range`.

chriselrod (approved):

Tokazama:
> > @chriselrod , I think the only issue you might run into with this one is that `_pick_range` is gone, so I put a deprecation in for that so it would be mildly annoying instead of breaking.
> 
> StrideArrays.jl uses it, but none of the other repos that I saw. It also depends on Static.jl, so the same PR bumping that can remove the use of `ArrayInterface._pick_range`.

Got it. I'll remove the deprecation, push this through, and then take a look at StrideArrays.jl

PR #373, opened 2023-02-17
Redo ArrayInterface to use extension packages

ChrisRackauckas:
* ArrayInterfaceCore -> ArrayInterface
* ArrayInterface -> ArrayInterfaceStaticExt
* All lib packages are extensions


test/static/indexing.jl
```diff
@@ -189,7 +189,13 @@ end
     end
 
     x = LinearIndices(map(Base.Slice, (static(0):static(3), static(3):static(5), static(-2):static(0))))
-    @test @inferred(ArrayInterface.static_getindex(x, 0, 3, -2)) === 1
+    
+    if VERSION >= v"1.7"
+        @test @inferred(ArrayInterface.static_getindex(x, 0, 3, -2)) === 1
+    else
+        @test ArrayInterface.static_getindex(x, 0, 3, -2) === 1
+    end
```
ChrisRackauckas, line 197: @chriselrod let me know if this needs and actual fix. This is the only thing really lost, and only on v1.6

PR #387, opened 2023-03-06
correct Requires usage to fix issues with PackageCompiler sysimages

KristofferC:
Basically, the rule for Requires to work with sysimages is that it should load no package non-relatively.

### DIFF
--- a/ext/ArrayInterfaceBandedMatricesExt.jl
+++ b/ext/ArrayInterfaceBandedMatricesExt.jl
@@ -1,8 +1,16 @@
 module ArrayInterfaceBandedMatricesExt
 
-using ArrayInterface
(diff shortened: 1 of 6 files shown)


ChrisRackauckas (approved):
Thanks!

KristofferC:
So `Adapt` is not loaded in ArrayInterface which means the `using ..Adapt` for Requires doesn't work. I don't see a way to be relocatable for Requries with that setup. I put it back how things were for that specific extension though.

ChrisRackauckas:
Adapt's stuff got upstreamed to Adapt.jl so that's fine.

KristofferC:
But there is a function below using it?

ChrisRackauckas:
🤦

Issue #404, opened 2023-03-28
`restructure` does not promote eltype for StaticArrays


gaurav-arya:
Is the following expected?
```
julia> using ArrayInterface, StaticArrays

julia> ArrayInterface.restructure(SA[1,2], SA[3.2, 4.5])
ERROR: InexactError: Int64(3.2)
Stacktrace:
 [1] Int64
   @ ./float.jl:900 [inlined]
 [2] convert
   @ ./number.jl:7 [inlined]
 [3] macro expansion
   @ ~/.julia/packages/StaticArraysCore/U2Z1K/src/StaticArraysCore.jl:81 [inlined]
 [4] convert_ntuple
   @ ~/.julia/packages/StaticArraysCore/U2Z1K/src/StaticArraysCore.jl:77 [inlined]
 [5] SArray
   @ ~/.julia/packages/StaticArraysCore/U2Z1K/src/StaticArraysCore.jl:113 [inlined]
 [6] SArray
   @ ~/.julia/packages/StaticArraysCore/U2Z1K/src/StaticArraysCore.jl:117 [inlined]
 [7] StaticArray
   @ ~/.julia/packages/StaticArrays/a4r2v/src/convert.jl:167 [inlined]
 [8] restructure(x::SVector{2, Int64}, y::SVector{2, Float64})
   @ ArrayInterfaceStaticArraysCoreExt ~/.julia/packages/ArrayInterface/xqHUD/ext/ArrayInterfaceStaticArraysCoreExt.jl:34
 [9] top-level scope
   @ REPL[5]:1
```

gaurav-arya:
Resolved by #405

PR #405, opened 2023-03-28
Expertiment with fixing static array restructure method

gaurav-arya:
The `restructure(x,y)` method for static arrays does not preserve the type of `y`, which is inconsistent with other methods of `restructure`. However, it may be that downstream packages rely on this behaviour. See #404, https://github.com/JuliaDiff/SparseDiffTools.jl/issues/237, https://github.com/JuliaArrays/ArrayInterface.jl/commit/691acbfcd0fa0175faa8f269730636b56dd24c99


gaurav-arya:
@ChrisRackauckas would appreciate if you could run workflows here. This makes `restructure` consistent with its implementation for Array's, GPUArray's, etc., but unclear if there are any downstream effects

ChrisRackauckas (approved):

PR #411, opened 2023-05-13
migrate simplified dimnames

Tokazama:
This is an exrtremely simplified version of `dimnames` that was previously here. Also drops all `StaticSymbol` requirements.


Tokazama:
@ChrisRackauckas, does this follow the format you want for the docs?

PR #417, opened 2023-05-30
Fix cholesky_instance for symmetric sparse

ChrisRackauckas:


src/ArrayInterface.jl
```diff
@@ -476,7 +476,7 @@ cheaply.
 function cholesky_instance(A::Matrix{T}, pivot = LinearAlgebra.RowMaximum()) where {T}  
     return cholesky(similar(A, 0, 0), pivot, check = false)
```
chriselrod, line 477: I don't know about the dense case, but it normally isn't necessary to pivot cholesky factorizations.

PR #429, opened 2024-02-14
Fixed can_setindex for MArray

jecs:
`can_setindex` now evaluates to `true` for `Type{<:StaticArraysCore.MArray}`.

See [Issue 428](https://github.com/JuliaArrays/ArrayInterface.jl/issues/428).

@Tokazama


Tokazama:
Just for the sake of completeness could you put a test in "/test/staticarrayscore.jl"? And then bump the patch version from 7.7.0 to 7.7.1 so we can release a fixed version.

ChrisRackauckas (approved):

PR #430, opened 2024-03-04
Specialize Array of Structs to Struct of Array for ReverseDiff

ChrisRackauckas:


ChrisRackauckas:
@avik-pal this is pretty hacky, let me know if you know of a better way.

PR #431, opened 2024-03-07
Preserve the dimensions of ReverseDiff TrackedArray

avik-pal:
- Fixes reversediff flattening arrays
- Removes old code


avik-pal:
@ChrisRackauckas I am guessing the sign difference is coming from the diffeqcallbacks bug that you fixed and that hasn't been propagated to GaussAdjoint?

PR #443, opened 2024-06-03
chore: add `has_trivial_array_constructor`

DhairyaLGandhi:
We need to know if we can construct the parent type given some arguments. This adds a check for whether an `applicable` method exists. It currently doesn't check for unionalls, and I am not sure of how to check for them. In some cases we want to keep the concrete type:

```julia
julia> ca = ComponentVector((x = rand(3), y = rand(4),))
ComponentVector{Float64}(x = [0.6549137106381634, 0.37555505280294565, 0.8521039568665254], y = [0.40314196291239024, 0.35484725607638834, 0.6580528978034597, 0.10055508457632167])

julia> convert(typeof(ca), rand(7))
ComponentVector{Float64}(x = [0.2282005678229746, 0.5803510438411744, 0.734611658286954], y = [0.9325262248600694, 0.49927588841708803, 0.01698109918493529, 0.16806075729787862])
```

I also need to understand what the appropriate tests here would be.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -52,6 +52,7 @@ BandedMatrices = "aae01518-5342-5314-be14-df237901396f"
 BlockBandedMatrices = "ffab5731-97b5-5995-9138-79e8c1846df0"
 CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
 ChainRules = "082447d4-558c-5d27-93f4-14fc19e9eca2"
+ComponentArrays = "b0b7db55-cfe3-40fc-9ded-d10e2dbeff66"
 GPUArraysCore = "46192b85-c4d5-4398-a991-12ede77f4527"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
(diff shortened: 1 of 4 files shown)


Tokazama:
I like the direction this is going but I'm not sure how widely applicable this will be given how the appropriate constructor method for may not always be the type itself. For example, if you have a reshaped array you generally use `reshape` and `view` if dealing with `SubArray`.

As far as tests, we just need to demonstrate that it works consistently for appropriate method definitions. In this case I think you'd just want to try the reconstruction in tests.

Should also have docs to make sure we can appropriately maintain and explain what's going on here.

DhairyaLGandhi:
Agreed, its related to how we want to use this trait within SciMLStructures as well. Maybe a slightly more generalized version of this question could be "can we convert to given type given certain arguments". In which case we can check for 

```julia
applicable(convert, T, args...)
```

ChrisRackauckas (approved):

ChrisRackauckas:
Doc build failing from this.

PR #451, opened 2024-07-30
revert #449

oscardssmith:
Since the SArray types are created in StaticArraysCore, unfortunately we need the extension to be on the library or else generated functions in 3rd party packages can miss the added methods. Furthermore, this breaks the ability to optimize `lu` since `StaticArraysCore` doesn't define `StaticArrays.LU` so this just is really annoying. I really think we should delete StaticArraysCore and just make it an alias for StaticArrays now that extension packages exist.

### DIFF
--- a/ext/ArrayInterfaceStaticArraysCoreExt.jl
+++ b/ext/ArrayInterfaceStaticArraysCoreExt.jl
@@ -21,10 +21,6 @@ ArrayInterface.can_setindex(::Type{<:StaticArray}) = false
 ArrayInterface.can_setindex(::Type{<:MArray}) = true
 ArrayInterface.buffer(A::Union{SArray, MArray}) = getfield(A, :data)
 
-function ArrayInterface.lu_instance(A::SMatrix{N,N}) where {N}
-    LU(LowerTriangular(A), UpperTriangular(A), SVector{N}(1:N))
-end
-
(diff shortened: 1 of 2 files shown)


ext/ArrayInterfaceStaticArraysCoreExt.jl
```diff
@@ -21,10 +21,6 @@ ArrayInterface.can_setindex(::Type{<:StaticArray}) = false
 ArrayInterface.can_setindex(::Type{<:MArray}) = true
 ArrayInterface.buffer(A::Union{SArray, MArray}) = getfield(A, :data)
 
-function ArrayInterface.lu_instance(A::SMatrix{N,N}) where {N}
```
ChrisRackauckas, line 24: why was this removed?

oscardssmith: because `LU` is defined in `StaticArrays` not `StaticArraysCore`

ChrisRackauckas: Didn't we need this for performance? What about having a StaticArraysExt for just this one?

oscardssmith: I tried that locally and it causes invalidations during precompile which is illegal. I'll just hack my way around it...

PR #461, opened 2025-05-06
Fully qualify constructor extension

jpthiele:
Extending constructors without module specification can lead to undefined behaviour in case of name clashes.  This is fixed by fully qualifying `ArrayInterface.BandedMatrixIndex` when it is extended in the banded matrix ext.

### DIFF
--- a/ext/ArrayInterfaceBandedMatricesExt.jl
+++ b/ext/ArrayInterfaceBandedMatricesExt.jl
@@ -46,7 +46,7 @@ function _bandsize(bandind, rowsize, colsize)
     end
 end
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Thanks! It would be good to test this better but it's such an edge case to hit.

jpthiele:
> Thanks! It would be good to test this better but it's such an edge case to hit.

Is there a reasonable way to catch and save warnings thrown by a function call without interrupting/stopping it?
If yes I could add something to [DownstreamTester](github.com/jpthiele/DownstreamTester.jl) since `Pkg.add()` already throws these warnings in nightly.

